PR #3, opened 2020-05-27
Add contributor docs and an example import using Template MCF + StatVarObservation.

tjann:
Note: Scripting Instance MCF might make more sense for data that needs to be refreshed daily. In this case, we could use the preprocessing script as the download process. So the preprocessing would need cron-ing.


docs/mcf_format.md
```diff
@@ -0,0 +1,321 @@
...
+hasResponseOption: E:SomeDataset->E1
+typeOf: SurveyItem
+dcid: C:SomeDataset->SurveyItem_Dcid
+```
+
+Where E is short for "Entity" and "C" is short for "Column".
+
+Contributors need not worry about the process of using template MCF to convert
+datasets to instance MCF, but roughly, for each row of the dataset, we create
+all the nodes in the template MCF specification, subbing in values from the
```
pradh, line 319: s/subbing in/substituting/

scripts/covid_tracking_project/historic_state_data/preprocess_csv.py
```diff
@@ -0,0 +1,60 @@
...
+      writer.writerow(processed_dict)
+
+# Automate Template MCF generation since there are 18 Statitical Variables.
+
+TEMPLATE_MCF_TEMPLATE = """
+Node: E:COVIDTracking_States->E{index}
+typeOf: StatVarObservation
+variableMeasured: {stat_var}
+observationAbout: C:COVIDTracking_States->GeoId
+observationDate: C:COVIDTracking_States->Date
```
pradh, line 53: quotes?

scripts/covid_tracking_project/historic_state_data/preprocess_csv.py
```diff
@@ -0,0 +1,60 @@
+import csv
```
pradh, line 1: (later is fine) might be good if the example has a test. that way, there is some chance that contributors will provide one.

docs/representing_statistics.md
```diff
@@ -0,0 +1,127 @@
...
+observationAbout: dcid:geoId/4865000
+observationDate: "2014"
+value: 39.4
+```
+
+Where [`MedianAge`](https://browser.datacommons.org/kg?dcid=MedianAge)
+is a `StatisticalVariable` schema node that only needs to be defined once.
+
+
+## Mapping `StatisticalVariable` and `StatVarObservation` to `StatisticalPopulation` and `Observation`
```
pradh, line 62: The content here sort presupposes user understands StatPop/Obs. Should we link in the data-model doc that delves into those? Like the one that I think Antares curated.

docs/representing_statistics.md
```diff
@@ -0,0 +1,127 @@
...
+Node: StatisticalPopulation_People_SanAntonio_TX
+typeOf: dcid:StatisticalPopulation
+populationType: dcid:Person
+location: dcid:geoId/4865000
+```
+
+The `Observation` copies the `StatisticalVariable`'s `measuredProperty`,
+`measurementMethod`, `measurementQualifier`, `unit`, etc.
+and the `StatVarObservation`'s `observationDate`. It also extracts the
+`statType`value from the `StatisticalVariable` and the `value` value from
```
edumorlom, line 95: should there be space after `statType` and before value?

docs/representing_statistics.md
```diff
@@ -0,0 +1,127 @@
...
+[Schema.org Data Model](https://schema.org/docs/datamodel.html),
+and its schema is a superset of
+[Schema.org schema](https://schema.org/docs/schemas.html).
+
+The most commonly used high-level schema types for representing statistical data are:
+
+- [`Class`](https://schema.org/Class)
+- [`Property`](https://schema.org/Property)
+- [`StatisticalPopulation`](https://schema.org/StatisticalPopulation)
+- [`Observation`](https://schema.org/`Observation`)
```
edumorlom, line 17: I think you didn't mean to put the ` in the url. I'm getting a 404 error

docs/representing_statistics.md
```diff
@@ -0,0 +1,127 @@
...
+typeOf: dcid:StatisticalPopulation
+populationType: dcid:Person
+location: dcid:geoId/4865000
+```
+
+The `Observation` copies the `StatisticalVariable`'s `measuredProperty`,
+`measurementMethod`, `measurementQualifier`, `unit`, etc.
+and the `StatVarObservation`'s `observationDate`. It also extracts the
+`statType`value from the `StatisticalVariable` and the `value` value from
+the `StatVarObservation` as its own `.*Value` property and value.
```
edumorlom, line 96: This last paragraph is a bit confusing, too many **values** in my opinion.

docs/mcf_format.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+To refer to an entity via its identifier (`id1`) in a namespace associated with
+a prefix 'foaf', we write 'foaf:id1'. To refer to an entity via its identifier
+in the local namespace, we use the prefix `l`.
+
+
+#### Special Namespaces
+
+The Data Commons graph is built on top of schema.org vocabulary and its Data
```
edumorlom, line 91: Sometimes you use Schema.org and other times schema.org

scripts/covid_tracking_project/historic_state_data/preprocess_csv.py
tjann: Seems table2mcf takes care of the quoting. Actually doesn't work if I add quotes to observationDate. I did add in dcs: to typeOf, variableMeasured.

docs/representing_statistics.md
tjann: Did a little shuffling, hopefully this helps.

docs/representing_statistics.md
tjann: I will do this and the test in a separate PR.

PR #24, opened 2020-06-12
Combining Python scripts for treasury rates import

intrepiditee:
Thanks for reviewing.

PR #25, opened 2020-06-12
Script downloads csv for treasury rates import

intrepiditee:
Only thing changed from #24 is the script.

Thanks for reviewing.


intrepiditee:
Close to do some cleanups

PR #30, opened 2020-06-16
Covid_tracking_states dataset

wh1210:


wh1210:
Thanks for your suggestions, that's pretty reasonable, I will keep this in mind and fix furtherly.

PR #32, opened 2020-06-16
move measurementMethod to StatVarObservation

intrepiditee:
Thanks for reviewing.


intrepiditee:
I though we would need one StatVar for one time series, no?

intrepiditee:
> With the mmethod change, it looks like we can consolidate a lot of the StatVars:
> 
> We'll end up with 3, I believe: one each for Treasury Bill, Note, and Bond. The names no longer need to refer to the mmethod.
> 
> Also, Ian is currently working on renaming our existing StatVars to this format: mprop_poptype_v1_v2... can we follow that here? E.g. InterestRate_TreasuryBond
I though we would need one StatVar for one time series, no?

intrepiditee:
> With the mmethod change, it looks like we can consolidate a lot of the StatVars:
> 
> We'll end up with 3, I believe: one each for Treasury Bill, Note, and Bond. The names no longer need to refer to the mmethod.
> 
> Also, Ian is currently working on renaming our existing StatVars to this format: mprop_poptype_v1_v2... can we follow that here? E.g. InterestRate_TreasuryBond

I though we would need one StatVar for one time series, no?

intrepiditee:
closed. please see #42

PR #36, opened 2020-06-17
Create measurement_methods.mcf

tjann:


schema/measurement_methods.mcf
```diff
@@ -0,0 +1,18 @@
+# File started on Jun 17, 2020 to start consolidating measurement method nodes
+# in one place.
+#
```
pradh, line 3: nit: Have a more general heading to what it does. The start date can be found from version history :)

PR #42, opened 2020-06-19
treasury rates schema changes

intrepiditee:
Thanks for reviewing.


scripts/us_fed/treasury_constant_maturity_rates/treasury_constant_maturity_rates.mcf
```diff
@@ -1,99 +1,88 @@
-Node: dcid:US_Treasury_1_Month_Constant_Maturity_Rate
-name: "US_Treasury_1_Month_Constant_Maturity_Rate"
+Node: dcid:InterestRate_TreasuryBill_1Month
+name: "InterestRate_TreasuryBill_1Month"
 typeOf: dcs:StatisticalVariable
-populationType: dcs:TreasuryBill
 measuredProperty: dcs:interestRate
+populationType: dcs:TreasuryBill
+maturity: 1Month
```
tjann, line 6: For quantitative values, we like to represent them in MCF as [Month 1] or [1 Month]. Either is ok. This is called a QuantityRange; here is an example of what happens when we specify using QuantityRange: https://browser.datacommons.org/kg?dcid=dc%2Fp%2F0wjznyrfjqls5

tjann: Unfortunately, that means the dcid of the QR will look like Month1 instead of 1Month (less pretty for the StatVar), but worth the tradeoff. Can think about whether we want to change our QR dcid generation to put the unit last. That will be a big change requiring reimporting a lot of big datasets, though. @pradh

pradh: Does improve readability, but it will be a churn-ful change.

intrepiditee: Fixed.

PR #45, opened 2020-06-19
Add demElevation node.

Spaceenter:
Used as a predicate in GeoNames import.

### DIFF
--- a/schema/dcschema.mcf
+++ b/schema/dcschema.mcf
@@ -2230,4 +2230,4 @@ Node: dcid:digitalElevationModel
 typeOf: schema:Property
 name: "digitalElevationModel"
 descriptionUrl: "http://download.geonames.org/export/dump/readme.txt"
-description: "Average elevation based on Digital Elevation Model (DEM), in meters, integer."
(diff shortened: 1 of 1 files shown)


schema/dcschema.mcf
```diff
@@ -2226,3 +2226,8 @@ name: "MedicalTest"
 typeOf: schema:Class
 subClassOf: dcs:MedicalProcedure
 
+Node: dcid:demElevation
+typeOf: schema:Property
+name: "demElevation"
+descriptionUrl: "http://download.geonames.org/export/dump/readme.txt"
+description: "Average elevation based on Digital Elevation Model (DEM), in meters, integer."
```
mvashishtha, line 2233: Should the DEM model, "average", and "unit" be part of the measurement or the variable, instead of property? Maybe just use https://schema.org/elevation instead?

Spaceenter: This is actually different from "elevation", see here http://download.geonames.org/export/dump/readme.txt  . Search for "elevation" and "dem". Seems like 'average' is the definition of this property?
I agree unit (meters) should go into observation -- changed.

PR #61, opened 2020-06-25
Fix typos in enum names

spiekos:


spiekos:
@googlebot I signed it!

PR #70, opened 2020-06-26
Add one line into covid tracking states statvars mcf

wh1210:


schema/StatisticalVariable.mcf
```diff
@@ -179,3 +179,99 @@ medicalStatus: dcs:PatientDeceased
...
+measuredProperty: dcs:count
+medicalStatus: dcs:PatientOnVentilator
+
+Node: dcid:CumulativeCount_MedicalConditionIncident_COVID_19_PatientOnVentilator
+typeOf: dcs:StatisticalVariable
+populationType: dcs:MedicalConditionIncident
+medicalCondition: dcs:COVID_19
+statType: dcs:measuredValue
+measuredProperty: dcs:cumulativeCount
+medicalStatus: dcs:PatientOnVentilator
```
tjann, line 277: We decided not to add the new StatisticalVariables to the schema folder. These are all the same as the ones in your import subdir, correct? If so, we can just undo these changes.

scripts/covid_tracking_project/historic_state_data/config.txt
```diff
@@ -3,12 +3,12 @@ import: {
...
+    mcf_url: "gs://resolved_mcf/external_tables/covid_tracking_states_new/nodes.mcf-00000-of-00001"
     # First date of imported data.
     start_date: "2020-01-22"
     # Last date of imported data.
-    end_date: "2020-06-22"
+    end_date: "2020-06-25"
     # Source data release schedule.
     update_frequency: "P1D"
-    next_release_date: "2020-06-23"
+    next_release_date: "2020-06-26"
```
tjann, line 13: Can also remove end_date and next_release_date here so that future no need to keep updating this.

scripts/covid_tracking_project/historic_state_data/preprocess_csv.py
```diff
@@ -63,7 +63,7 @@
 variableMeasured: dcs:{stat_var}
 measurementMethod: dcs:CovidTrackingProject
 observationAbout: C:COVIDTracking_States->GeoId
-observationDate: C:COVIDTracking_States->Dates
+observationDate: C:COVIDTracking_States->Date
```
tjann, line 66: Oof, good catch. How did this work before?

wh1210: No worry about this, I am thinking this is because I unintentionally add a 's' when using 'ctrl + s' to save file. The tmcf in prod is the correct one~~

schema/StatisticalVariable.mcf
wh1210: Yes, let me do some changes.

scripts/covid_tracking_project/historic_state_data/config.txt
wh1210: Got it.

schema/StatisticalVariable.mcf
wh1210: Yep, this is also what I am thinking as this will cause conflicts easily and would potentially cause some problems unintentionally.

PR #86, opened 2020-07-02
Fix spelling error

spiekos:


spiekos:
@googlebot I signed it!

PR #89, opened 2020-07-02
Import automation: Cloud Build scripts

intrepiditee:
This PR includes scripts that would be run by Cloud Build on commits. task.sh is a bash script that installs and runs pip to install dependencies and runs create_task.py to create a Cloud Tasks task that passes information about the commit to the executor.

The scripts have been tested on Cloud Build.

### DIFF
--- a/import-automation/cloudbuild/cloudbuild.yaml
+++ b/import-automation/cloudbuild/cloudbuild.yaml
@@ -1,16 +1,16 @@
 steps:
 - name: 'gcr.io/cloud-builders/gcloud'
   entrypoint: 'bash'
-  args: ['import-automation/task.sh']
+  args: ['import-automation/cloudbuild/task.sh']
   env:
-  - 'COMMIT_SHA=$COMMIT_SHA',
-  - 'REPO_NAME=$REPO_NAME',
-  - 'BRANCH_NAME=$BRANCH_NAME',
-  - 'HEAD_BRANCH=$_HEAD_BRANCH',
-  - 'BASE_BRANCH=$_BASE_BRANCH',
+  - 'COMMIT_SHA=$COMMIT_SHA'
+  - 'REPO_NAME=$REPO_NAME'
+  - 'BRANCH_NAME=$BRANCH_NAME'
+  - 'HEAD_BRANCH=$_HEAD_BRANCH'
+  - 'BASE_BRANCH=$_BASE_BRANCH'
   - 'PR_NUMBER=$_PR_NUMBER'
   - 'TASK_LOCATION_ID=$_TASK_LOCATION_ID'
   - 'TASK_QUEUE_NAME=$_TASK_QUEUE_NAME'
   - 'TASK_PROJECT_ID=$_TASK_PROJECT_ID'
-  - 'HANDLER_ENDPOINT=$_HANDLER_ENDPOINT'
+  - 'HANDLER_SERVICE=$_HANDLER_SERVICE'
   - 'HANDLER_URI=$_HANDLER_URI'
--- a/import-automation/cloudbuild/cloudbuild.yaml
+++ b/import-automation/cloudbuild/cloudbuild.yaml
@@ -1,0 +1,16 @@
+steps:
+- name: 'gcr.io/cloud-builders/gcloud'
+  entrypoint: 'bash'
+  args: ['import-automation/task.sh']
+  env:
+  - 'COMMIT_SHA=$COMMIT_SHA',
+  - 'REPO_NAME=$REPO_NAME',
+  - 'BRANCH_NAME=$BRANCH_NAME',
+  - 'HEAD_BRANCH=$_HEAD_BRANCH',
+  - 'BASE_BRANCH=$_BASE_BRANCH',
+  - 'PR_NUMBER=$_PR_NUMBER'
+  - 'TASK_LOCATION_ID=$_TASK_LOCATION_ID'
+  - 'TASK_QUEUE_NAME=$_TASK_QUEUE_NAME'
+  - 'TASK_PROJECT_ID=$_TASK_PROJECT_ID'
+  - 'HANDLER_ENDPOINT=$_HANDLER_ENDPOINT'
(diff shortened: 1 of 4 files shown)


import-automation/cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,16 @@
+steps:
+- name: 'gcr.io/cloud-builders/gcloud'
```
jeffreyoldham, line 2: Perhaps add a comment explaining what this chunk of code does.  Your pull request description is probably a superset of what to write here.

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+    in TASK_BODY_FIELDS in the environment variables.
+
+    PR_NUMBER is converted to an integer.
+
+    Returns:
+        The task body as a dict.
+    """
+    task_body = {}
+    for field in TASK_BODY_FIELDS:
+        task_body[field] = os.environ[field]
```
jeffreyoldham, line 49: What happens when a particular field is not in os.environ?  Should get() be used?

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Args:
+        task_body: Request body of the task as a dict that
+            will be serialized into json.
+        project_id: ID of the Google Cloud project as a string.
+        location_id: ID of the location where the task queue is hosted.
+        queue_name: Name of the task queue.
+        service: Name of the App Engine service that will handle the task.
+        endpoint: Relative URL of the App Engine task handler endpoint.
+    """
```
jeffreyoldham, line 73: It would be nice to have a
Returns:
or "Effects:" line in the comment.

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+            },
+            'http_method': 'POST',
+            'relative_uri': endpoint,
+            'body': body.encode(),
+            'headers': {
+                'Content-Type': 'application/json'
+            }
+        }
+    }
+    client.create_task(parent, task)
```
jeffreyoldham, line 90: Should this return the client even if the caller does not use it?

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def main():
+    """
+    Creates a Cloud Tasks task that ships information about a GitHub commit
+    to the executor.
+    """
+    task_body = create_body()
```
jeffreyoldham, line 98: Can we just inline this value, avoiding the need for the use-once variable?

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+    create_task(
+        task_body=task_body,
+        project_id=os.environ['TASK_PROJECT_ID'],
```
jeffreyoldham, line 101: Similar question to above about how to handle the case where the environment does not have the specified values.

import-automation/cloudbuild/cloudbuild.yaml
```diff
@@ -0,0 +1,16 @@
...
+  entrypoint: 'bash'
+  args: ['import-automation/cloudbuild/task.sh']
+  env:
+  - 'COMMIT_SHA=$COMMIT_SHA'
```
jeffreyoldham, line 6: I worry about using environment variables to pass around values.  Explicitly using command-line arguments is better because it is explicit.  They are also type checked and have default values.  However, sometimes they are necessary in some systems.

intrepiditee: The main reason I chose to use environmental variables was that I did not want to add another dependency as Cloud Build would have to install it every time.

import-automation/cloudbuild/create_task.py
intrepiditee: These fields are required. If a field is missing, the script will fail.

import-automation/cloudbuild/create_task.py
```diff
@@ -0,0 +1,109 @@
...
+
+    PR_NUMBER is converted to an integer.
+
+    Returns:
+        The task body as a dict.
+    """
+    task_body = {}
+    for field in TASK_BODY_FIELDS:
+        task_body[field] = os.environ[field]
+    task_body['PR_NUMBER'] = int(task_body['PR_NUMBER'])
```
rsned, line 50: can this conversion fail? this is a pull request # not the git PR hash number right?

intrepiditee: Yes it is the pull request number. The conversion should not fail.

import-automation/cloudbuild/create_task.py
intrepiditee: fixed

import-automation/cloudbuild/create_task.py
intrepiditee: fixed

import-automation/cloudbuild/cloudbuild.yaml
intrepiditee: added

import-automation/cloudbuild/create_task.py
intrepiditee: added

import-automation/cloudbuild/create_task.py
jeffreyoldham: Let's write a comment to that effect, making it explicit.

import-automation/cloudbuild/cloudbuild.yaml
jeffreyoldham: It is still worrisome, but I defer.

import-automation/cloudbuild/create_task.py
intrepiditee: added

import-automation/cloudbuild/create_task.py
intrepiditee: added a comment to state this

PR #91, opened 2020-07-07
This library implements the start of a sharded writer in Go.

rsned:
This library implements the start of a sharded writer in Go to generate.files of the desired size for Data Commons imports (100MB).
Adds a variety of tests and test helper methods to make sure a wide selection of test cases can be comprehensively covered.


rsned:
Comments and feedback were in [PR #48](https://github.com/datacommonsorg/data/pull/48) before my repo got too tangled. This PR is of the final product of those comments:

PR #94, opened 2020-07-08
Update COVID Tracking data (schema change and data refresh)

wh1210:
https://github.com/datacommonsorg/data/pull/93 This is the subsequent step for this PR.

1. Update the data for covid tracking project both states and US level.
2. Update the config file to be corresponding to the new resolved mcf file position.(US is the same as before, so no need to change it.)

These 2 PRs were to fix the medicalCondition -> incidentType for MedicalConditionIncident StatVars.


tjann:
Can you link to your last PR, and indicate that this update was to fix the medicalCondition -> incidentType for MedicalConditionIncident StatVars?

PR #97, opened 2020-07-08
added eurostat processing script

jinpz:


scripts/eurostat/process_eurostat.py
```diff
@@ -0,0 +1,130 @@
...
+            raw_value = row_values[j + 1].strip()
+            value_with_note = raw_value.split(' ')  # want to distinguish between the ones with notes and without
+            if len(value_with_note) == 2:
+                value = value_with_note[0]
+                note = value_with_note[1]
+            elif len(value_with_note) == 1:
+                value = value_with_note[0]
+                note = ':'
+            else:
+                print(value_with_note)
```
jeffreyoldham, line 58: If we use the abseil logging module suggested below, we can logging.fatal("Entry should have a value and an optional note: %s", value_with_note).  This will produce the message and throw an exception, killing the program and producing a stack trace.

scripts/eurostat/process_eurostat.py
```diff
@@ -0,0 +1,130 @@
...
+
+def unzip_single(filename_in):
+    filename_out = filename_in[:-3]  # to trim the .gz
+    with gzip.open(filename_in, 'rb') as f_in:
+        with open(filename_out, 'wb') as f_out:
+            shutil.copyfileobj(f_in, f_out)
+    return filename_out
+
+
+def first_stage_processing(df, separation_symbol):
```
jeffreyoldham, line 15: Would a more descriptive function name help the reader?  Unfortunately, I am not sure what to suggest.

scripts/eurostat/process_eurostat.py
```diff
@@ -0,0 +1,130 @@
...
+    for v in variable_column:
+        if v != 'time' and v != time_invariant_var:
+            combined_variable_name = combined_variable_name + v + separation_symbol
+    combined_variable_name = combined_variable_name[:-len(separation_symbol)]  # remove last underscore
+    # insert a dummy variable for two dimensional dataset. Note most of the time this should not happen
+    if combined_variable_name == '':
+        combined_variable_name = 'dummy_var_column'
+    total_record_count = len(df)*len(time_columns)
+    all_records = []
+    for i in range(len(df)):
```
jeffreyoldham, line 39: Do we want to process the header row?  I think so, but I just want to ask to be sure.

jinpz:
Thanks Jeffrey for the review. I agree that the processing script should have sufficient comments. I will share with you a google drive folder to see what input and output look like.

PR #105, opened 2020-07-10
Added new schema for WorldBank priority 0 import

IanCostello:
Added new schema for the first set of the large scale WorldBank import.


tjann:
Thanks, Ian. Due to sync problems, we're just adding schema to g3 now. I've sent out a CL and it should be in soon.

PR #117, opened 2020-07-15
Minor fixes to covid_mobility and add manifest.json

pradh:
+ Fix typo in test (previously it was broken)
+ Delete unnecessary large CSV file
+ Add manifest.json


scripts/google/covid_mobility/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+    "import_specifications": [
+        {
+            "import_name": "data",
```
tjann, line 4: Is the import name intentionally vague?

pradh:
Thanks Eduardo!

scripts/google/covid_mobility/manifest.json
```diff
@@ -0,0 +1,20 @@
...
+            "curator_emails": [
+                "edumorales@google.com"
+            ],
+            "provenance_url": "https://www.google.com/covid19/mobility",
```
intrepiditee, line 8: Could we also add a "provenance_description"? It is a required field in our design.

PR #129, opened 2020-07-21
Update covidmobility.py

edumorlom:
We were skipping the population node if the first row had had no value.
Now we always write it on the first iteration, then we check to see if there is a value.


edumorlom:
@pradh I will do a separate pull request to update the unit test. Wanted to get this in asap.

PR #142, opened 2020-07-22
Import automation: replace Python script with gcloud command

intrepiditee:
Just found out that we didn't actually need Python. This halves the Cloud Build setup time (14 seconds now).


scripts/covid_tracking_project/historic_state_data/manifest.json
```diff
@@ -6,6 +6,7 @@
                 "hhhhhwu@google.com"
             ],
             "provenance_url": "https://www.covidtracking.com",
+            "provenance_description": "The COVID Tracking Project is a volunteer organization launched from The Atlantic and dedicated to collecting and publishing the data required to understand the COVID-19 outbreak in the United States.",
```
jeffreyoldham, line 9: I am not sure this should be part of this pull request.

PR #143, opened 2020-07-22
Format to Google style.  No semantic changes are expected.

jeffreyoldham:
Command: yapf --style google --recursive --in-place .


jeffreyoldham:
@rsned @shijunjie This is a replacement for PR 136.  Some of the changes are degradations, but it is probably better to have consistency rather than fix a handful of suboptimalities.

PR #146, opened 2020-07-22
added cause of death import, processed csv are too large for now

jinpz:
Tiffany wants to take a look at the enums before manifesting it


tjann:
Birth Death Migration is checked in here: https://github.com/datacommonsorg/data/tree/master/scripts/eurostat/birth_death_migration

The Death by Cause needed some more work but probably won't be attended to in near future, so I've closed this PR.

PR #151, opened 2020-07-23
Ensure "df" is not marked as an unacceptable variable name.

jeffreyoldham:


jeffreyoldham:
@rsned Will you please review?

PR #161, opened 2020-07-24
Add top-level requirements.txt.

jeffreyoldham:
Eventually this will be the one requirements.txt file in the scripts/
directory tree.

This list is the union of existing uses minus Python standard
libraries.  Version numbers were grabbed from some existing
requirements.txt files.

This may break some Python code which does not have their own
requirements.txt.  However, how they correctly import dependences is
already in question.

The next steps include:
1. Ensure all existing "unprotected" code compiles and runs using this
   requirements.txt file.
2. Remove lower-level requirements.txt files.


jeffreyoldham:
@rsned Is this a reasonable approach?

tjann:
Drive by comment: I remember when we were reviewing Jay's design, oldham@ and shanth@ both said requirements.txt should be import-directory specific since different imports may use different versions of libraries? Just curious, what new understandings are motivating this change?

rsned:
LGTM

PR #163, opened 2020-07-24
 fix up files that failed the python presubmit

rsned:
these got re-formatted with local edits. presubmit showed them, but didnt block submit.  re-applied the yapf.


scripts/covid_tracking_project/historic_us_data/preprocess_csv.py
```diff
@@ -41,18 +41,15 @@
                 ('CumulativeCount_MedicalConditionIncident'
                  '_COVID_19_PatientRecovered'):
                     row_dict['recovered'],
-                ('CumulativeCount_MedicalConditionIncident',
-                 '_COVID_19_PatientDeceased'):
+                ('CumulativeCount_MedicalConditionIncident', '_COVID_19_PatientDeceased'):
```
jeffreyoldham, line 44: If these fit on a single line, we might as well concatenate them.  Here and below.

scripts/us_bea/states_gdp/test_import.py
```diff
@@ -18,10 +18,12 @@
...
     python3 test_import.py
 """
+
 import unittest
+
 import pandas as pd
-import import_data
-import import_industry_data_and_gen_mcf
+from us_bea.states_gdp import import_data
+from us_bea.states_gdp import import_industry_data_and_gen_mcf
```
jeffreyoldham, line 26: I think these should be

import us_bea.states_gdp

and change all uses to "states_gdp.import_data" and "states_gdp.import_industry_data_and_gen_mcf".

scripts/covid_tracking_project/historic_us_data/preprocess_csv.py
rsned: This is how the formatter spit things out.

scripts/us_bea/states_gdp/test_import.py
rsned: We should add this to the best practices notes as well since we are likely to have many top level scripts paths with bushy directory trees and not just one 1:1.

jeffreyoldham: Does https://google.github.io/styleguide/pyguide.html#22-imports suffice?

PR #171, opened 2020-07-29
Remove lower-level requirements.txt files.  Fix code.

jeffreyoldham:
This is the second step of a two-part change of migrating to a
top-level requirements.txt file.  All codes with its own
requirements.txt files were run and results checked with and without
their local versions for being the same.

One program needed modification to execute.


jeffreyoldham:
@rsned Would you be willing to review?

PR #188, opened 2020-08-04
More Unit test + adds check for column number

edumorlom:
@pradh as promised, here is the unit test for that bug that left some nodes unresolved.

Fixes issue with extra columns. Now it ignores if != 14 columns.


scripts/google/covid_mobility/covidmobility.py
```diff
@@ -68,16 +68,25 @@ def csv_to_mcf(input_path: str, output_path: str) -> None:
...
+    # Input file.
     f_input = open(input_path, 'r')
+    # Output file.
     f_output = open(output_path, 'a+')
 
     csv_reader: DictReader = DictReader(f_input)
 
     for row in csv_reader:
+        # When this script was written, there were 14 rows.
+        # If there aren't exactly 14 rows, continue.
```
pradh, line 80: Will it then silently not process anything? Should we fail loudly instead, so we can fix the script?

edumorlom: Now i throw an exception with a human readable message. Fixed.

PR #192, opened 2020-08-06
Make cloudbuild for go work; fix lint errors in place_name_resolver.

tjann:


tjann:
@pradh 
Yeah, the cloudbuild runs `go test ./...`, and what that looks like locally seems to reasonable: 

image

tjann:
Turning on requirement for python and go checks for all PRs: 
image
@beets

PR #206, opened 2020-08-10
Add action to auto-approve copybara-service bot.

tjann:


.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,13 @@
+name: Auto approve
```
beets, line 1: please make the name more specific (auto approve copybara)

tjann:
feedback/copybara and import/copybara won't run bc these files do not affect files tracked by copybara. I do not know why they are labeled as `Expected -- Waiting for status to be reported` I will use admin privileges to merge, and ask around to see if anyone has a solution for requiring these checks to run and pass, but *only when applicable to copybara tracked files*.

PR #237, opened 2020-08-17
Import automation: authenticating with the proxy

intrepiditee:
I added authentication to the proxy. This PR makes the executor authenticate with it.

Thanks for reviewing.

### DIFF
--- a/import-automation/executor/app/service/import_service.py
+++ b/import-automation/executor/app/service/import_service.py
@@ -25,6 +25,7 @@ import requests
 from google.cloud import storage
 
 from app.executor import import_target
+from app.service import iap_request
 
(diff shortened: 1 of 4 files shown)


import-automation/executor/app/service/import_service.py
```diff
@@ -126,6 +127,8 @@ def __init__(self, project_id: str, unresolved_mcf_bucket_name: str,
                 unresolved_mcf_bucket_name bucket, as a string.
             executor_output_prefix: Output prefix in the
                 resolved_mcf_bucket_name bucket, as a string.
+            client_id: Oauth client ID to authenticate with the proxy,
```
jeffreyoldham, line 130: The typical spelling is "OAuth".

PR #245, opened 2020-08-18
Make OECD changes for better resolution

tjann:
1. Append "AdministrativeArea1" to the names of TL2 geos. This is to improve results of place_name_resolver.
2. Move special casing logic (do not use resolver output) from utils (called per-script) to the `clean_geos_resolved_to_dict.py` script.
3. Added US States to special casing logic.

Also had to change some things in order to get it to run.
1. Decrease resolver batch size to 50, with a 1 second sleep. This is to satisfy 50 QPS limit on geocoding API.
2. Remove  `quoting=csv.QUOTE_NONE` from `to_csv` (I was getting errors from running without this change).


tools/place_name_resolver/resolver.go
```diff
@@ -236,6 +238,7 @@ func geocodePlaces(mapCli MapsClient, placeId2Dcid *map[string]string, tinfo *ta
 			go geocodeOneRow(j, placeId2Dcid, tinfo, mapCli, &wg)
 		}
 		wg.Wait()
+		time.Sleep(1 * time.Second)
```
pradh, line 241: curious why is this needed? (add a comment)

tjann: Added a comment.

scripts/oecd/regional_demography/deaths/preprocess_csv.py
```diff
@@ -14,21 +14,22 @@
...
 
 # Read csv from source csv
 df = pd.read_csv('REGION_DEMOGR_death_tl3.csv')
 df = df[['TL', 'REG_ID', 'Region', 'VAR', 'SEX', 'Year', 'Value']]
+
 # First remove geos with names that we don't have mappings to dcid for.
 regid2dcid = dict(json.loads(open('../regid2dcid.json').read()))
-nuts = dict(json.loads(open('../region_nuts_codes.json').read()))
-df = df[df['REG_ID'].isin(nuts.keys()) | df['REG_ID'].isin(regid2dcid.keys())]
+df = df[df['REG_ID'].isin(regid2dcid.keys())]
```
wh1210, line 28: Will this line lose many data if not caught by regid2dcid.json?

wh1210: Sorry, can ignore this!

tjann:
> Not a full-review, just a drive-by comment.
> 
> Might be interesting to see some diff stats on the resolution results, before and after the fix.

Here are some stats for the change:

**Using the same placeid2dcid map:**

## Unresolved places (unable to get DCID):
- Before: 427
- After: 371

### Due to empty geocoding
- Before: 37
- After: 39

### Due to missing dcid for place
- Before: 390
- After: 332

## Number of mapped OECD geos
- Before: 1983
- After: 2424

## OECD to DCID map deltas
### Both present
- Total: 2132
- Changed due to change: 175
### Both missing
- Total: 310
### Missing before
- Total: 117
### Missing after
- Total: 61
### Total diff (pandas considers nan != nan)
- Total: 663

### Sanity check
663 = 175+117+61 + 310

pradh:
`AT13           nuts/AT13  geoNamesId/2761367`
Ideally, nuts/AT13 and the geoNames entity should probably be resolved to one node. nuts/AT13 was the original DCID, and geoNamesId was chosen after this PR.

--- The name=>placeId change is right. The problem is the placeId=>dcid mapping. When we replace geonames with wikidata, I think this should be improved.

A regression example is 
`BE324          nuts/BE324  wikidataId/Q220145`
BE324 is an AdminLevel4 "Mouscron" and the wikidataId is AdminLevel3 "Arrondissment of Mouscron". Since this is AA4, the resolution is not affected by adding AdministrativeArea1. While since this is a NUTS code, it'll eventually be replaced with our special cases, this shows we possibly have degressions/randomness.

--- This is nondeterminism in name=>placeId mapping (aka geocoding). I don't think we can do much about it unless we special case.

```
OECD_ID     prod     rc
ITC47 dcid:wikidataId/Q6221 dcid:nuts/ITC47
RO125 dcid:iso/RO-MS dcid:nuts/RO125
UKE12 dcid:geoNamesId/2650345 dcid:nuts/UKE12
EL301 dcid:wikidataId/Q755165 dcid:nuts/EL301
FRK14 dcid:nuts/FR724 dcid:nuts/FRK14
FRE11 dcid:nuts/FR301 dcid:nuts/FRE11
PL637 dcid:wikidataId/Q572754 dcid:nuts/PL637
AT123 dcid:wikidataId/Q82500 dcid:nuts/AT123
UKH17 dcid:wikidataId/Q1854245 dcid:nuts/UKH17
EL526 dcid:wikidataId/Q206957 dcid:nuts/EL526
DE918 dcid:wikidataId/Q498575 dcid:nuts/DE918
HU331 dcid:iso/HU-BK dcid:nuts/HU331
ME20 dcid:wikidataId/Q131429 dcid:iso/MX-OAX  #went from some no-name place to the correct Oaxaca TL2.
FRY3 dcid:country/GUF dcid:nuts/FRY3
FRF33 dcid:nuts/FR413 dcid:nuts/FRF33
```

--- These generally seem ok?

PR #268, opened 2020-08-24
Fixing manifest for covid mobility

intrepiditee:
1. import_name changed to Google_CovidMobility
2. fixes cron_schedule: "30 11 * * *" means every day at 11:30 AM UTC, which converts to 4:30 AM pacific
3. adds provenance_description, which is required
4. adds test to scripts

@pradh This changes the import name, which I think affects the production pipeline. Could you review? Thanks.


intrepiditee:
@jeffreyoldham Could you change the import name in g3? Then we can merge.

jeffreyoldham (approved):
Thanks.

PR #282, opened 2020-08-29
Fixing/adding manifests for covid tracking and Italy covid

intrepiditee:
For ItalyPCMDPC_COVID19, according to https://github.com/pcm-dpc/COVID-19, the dataset is updated daily at 18:00 Italy time (Central European Summer Time), which converts to 4:00 PM UTC.

Thanks for reviewing.


intrepiditee:
@jeffreyoldham Could you change the import names for covid tracking in g3 and add the manifest for Italy covid? Then we can merge

PR #283, opened 2020-08-31
US States geojson simplification

fpernice-google:
US States geojson simplification written to MCF. 

@beets @pradh @jeffreyoldham for review.


scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -36,7 +37,7 @@ def __init__(self):
         self.simplifier = simplify.GeojsonSimplifier()
         self.simple_geojsons = {}
 
-    def generate_simple_geojsons(self, verbose=True):
+    def generate_simple_geojsons(self, verbose=True, show=False):
```
jeffreyoldham, line 40: If you got logging working, there is no need for this parameter.  Logging works without output to the stdout or stderr unless the user specifies e wants output.  However, logging is produced in some obscurely named files for inspection if a run failed.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -53,26 +54,41 @@ def generate_simple_geojsons(self, verbose=True):
...
         plt.show()
 
-    # def generate_mcf(self, path='low_res_geojsons.mcf'):
-    #     with open(path, 'w') as f:
+    def generate_mcf(self, path='low_res_geojsons.mcf'):
+        temp = (
+            "Node: dcid:{geoid}\n"
+            "typeOf: dcs:{type}\n"
+            "geoJsonCoordinatesLowRes: {coords_str}\n\n"
+        )
```
jeffreyoldham, line 79:
```suggestion
        temp = "\n".join([
            "Node: dcid:{geoid}",
            "typeOf: dcs:{type}",
            "geoJsonCoordinatesLowRes: {coords_str}\n"
        ])
```

I am not sure this is an improvement.  It does mean that the programmer need not worry about forgetting to add a newline.

Triple quotes section is an alternative:
temp = '''
Node: dcid:{geoid}
typeOf: dcs:{type}
geoJsonCoordinatesLowRes: {coords_str}

'''
That might be better.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -53,26 +54,41 @@ def generate_simple_geojsons(self, verbose=True):
...
+                geostr = geojson.dumps(self.simple_geojsons[geoid])
+                f.write(temp.format(geoid=geoid, type="State",
+                                    coords_str=geostr))
 
 
-if __name__ == "__main__":
+def main(_):
     gen = McfGenerator()
-    gen.generate_simple_geojsons()
+    gen.generate_simple_geojsons(show=True, verbose=True)
```
jeffreyoldham, line 89: `show` probably needs to be a command-line argument.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -53,26 +54,41 @@ def generate_simple_geojsons(self, verbose=True):
...
-                                          show=False)
+                                          show=show)
+                    self.simple_geojsons[geoid] = simple
             except AssertionError:
                 logging.error("Simplifier failure on GeoJSON below:\n", coords)
         plt.show()
 
-    # def generate_mcf(self, path='low_res_geojsons.mcf'):
-    #     with open(path, 'w') as f:
+    def generate_mcf(self, path='low_res_geojsons.mcf'):
```
jeffreyoldham, line 74: Let's add a function comment.

scripts/us_census/geojsons_low_res/plotter.py
```diff
@@ -26,24 +26,21 @@
...
-flags.DEFINE_string('simplified_path',
-                    default=None,
-                    help='Path to simplified geojson to be compared.')
-flags.mark_flag_as_required('original_path')
-flags.mark_flag_as_required('simplified_path')
-
 _, (ax1, ax2) = plt.subplots(ncols=2, sharex=True, sharey=True)
 
 
 def compare_plots(geojson1, geojson2, show=True):
```
jeffreyoldham, line 32: Please add a function comment per https://google.github.io/styleguide/pyguide.html#383-functions-and-methods.

scripts/us_census/geojsons_low_res/plotter.py
```diff
@@ -26,24 +26,21 @@
...
-    f2.set_title('Simplified.')
+    if show:
+        _, (new_ax1, new_ax2) = plt.subplots(ncols=2, sharex=True, sharey=True)
+        f1 = geojson1.plot(ax=new_ax1)
+        f2 = geojson2.plot(ax=new_ax2)
+        f1.set_title('Original.')
+        f2.set_title('Simplified.')
+    else:
+        f1 = geojson1.plot(ax=ax1)
+        f2 = geojson2.plot(ax=ax2)
```
jeffreyoldham, line 41: How are `ax1` and `ax2` defined?  Are they global variables?

scripts/us_census/geojsons_low_res/plotter.py
```diff
@@ -26,24 +26,21 @@
...
+        f1.set_title('Original.')
+        f2.set_title('Simplified.')
 
     if show:
         plt.show()
```
jeffreyoldham, line 46: Choice 1: Move plt.show() within the `if show` conditional.
Choice 2: Move the `set_title` out of the conditionals because they apply to both cases.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: Removed `verbose` param.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: Done.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: Changed to `.join` approach. Triple quotes leave a weird indent in all lines after the first.

scripts/us_census/geojsons_low_res/plotter.py
fpernice-google: Added.

scripts/us_census/geojsons_low_res/plotter.py
fpernice-google: They are global variables yes. This is needed in order to have multiple geographies drawn in one plot.

scripts/us_census/geojsons_low_res/plotter.py
fpernice-google: Option (1) done.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -46,33 +47,51 @@ def generate_simple_geojsons(self, verbose=True):
...
+        """Writes the simplified GeoJSONs to an MCF file.
 
+        Args:
+            path: Path to MCF file to write simplified GeoJSON to.
+        """
+        temp = "\n".join([
+            "Node: dcid:{geoid}",
+            "typeOf: dcs:{type}",
+            "geoJsonCoordinatesLowRes: {coords_str}",
```
beets, line 80: I wonder if LowRes is the right term here (it's still pretty high res, high enough for display at large sizes). I can't find some standard classification scheme, but found this: https://www.naturalearthdata.com/downloads/

There should be an additional node to define this property - will that be hand written separately for import? For geoJsonCoordinates, its:

```
Node: dcid:geoJsonCoordinates
typeOf: schema:Property
name: "geoJsonCoordinates"
description: "Geographical boundary of a place in GeoJSON format."
domainIncludes: schema:Place
rangeIncludes: schema:Text
```

beets:
please take a look at the failing tests too

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: @pradh suggested the name `geoJsonCoordinatesLowRes`. He also indicated that the creation of the new property had to be done by a full time DataCommons person.

pradh: RE: naming... we already have the "geoJsonCoordinates" property. I was just trying to pick an alternate property name. Any other reasonable name works too.

RE: who can add it... my comment that a DC person should add it was before tjann@ set up the schema directory. you can make a separate change to add it next to https://github.com/datacommonsorg/data/blob/f1a4f35019f87b577049ff39ca9fd9529d2f0bcd/schema/dcschema.mcf#L323-L328

tjann: @fpernice-google it might be worth coming up with a more specific name than something involving "low res". E.g. today, Guha said that you got 20% (or 20x) reduction, but he might want even more. To leave room for multiple levels, I would propose something more specific. Throwing some ideas out there, but lmk what you come up with: 

douglasPeuckerEpsilon0.01GeoJsonCoordinates
douglasPeucker0.01GeoJsonCoordinates
rdpEpsilon0.01GeoJsonCoordinates

Neither of these seem very idea since RDP is probably not commonly known. geoJsonCoordinatesDouglasPeuckerSimplified0.01?

pradh: good idea. very slightly prefer geoJsonCoordinates<foo> over <foo>GeoJsonCoordinates.

tjann:
Can you also update the PR description? @fpernice-google

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: See #284 for final schema.

fpernice-google:
> Can you also update the PR description? @fpernice-google

Done.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -48,12 +64,14 @@ def generate_simple_geojsons(self, show=False):
             if int(geoid.split('/')[1]) > 100:
                 continue
             logging.info(f"Simplifying {geoid}...")
+            if eps is not None:
+                self.eps = eps
```
jeffreyoldham, line 60: This side effect should be documented.

It might be worth considering whether the function should just use `eps` if given and otherwise `self.eps` without any side effect.  If a person wants the side effect, the person should make an explicit call.

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -66,32 +84,58 @@ def generate_simple_geojsons(self, show=False):
...
         Args:
             path: Path to MCF file to write simplified GeoJSON to.
+            mode: mode parameter to be passed to open() function when writing
+                  to MCF.
         """
         temp = "\n".join([
+<<<<<<< HEAD
+            "Node: dcid:{geoid}", "typeOf: dcs:{type}",
+            "geoJsonCoordinatesDP{level}: {coords_str}", "\n"
+=======
```
jeffreyoldham: Merge problems.

PR #288, opened 2020-09-01
Added compressed US counties

fpernice-google:
US county equivalent of #283 (for US states). DP1 is around 12X compression, DP2 is around 25X, and DP3 is around 35X.


scripts/us_census/geojsons_low_res/download.py
```diff
@@ -90,22 +94,32 @@ def download_data(self, place='country/USA'):
         # to get the 0th element explicitly.
         assert len(geolevel[place]) == 1
         geolevel = geolevel[place][0]
+
+        level_to_fetch = geolevel
```
jeffreyoldham, line 98: Can we just use `geolevel` instead of defining another variable?

scripts/us_census/geojsons_low_res/download.py
```diff
@@ -90,22 +94,32 @@ def download_data(self, place='country/USA'):
...
+        for i in range(level):
```
jeffreyoldham, line 99:
```suggestion
```

scripts/us_census/geojsons_low_res/download.py
```diff
@@ -90,22 +94,32 @@ def download_data(self, place='country/USA'):
...
+            if level_to_fetch not in self.LEVEL_MAP:
+                raise ValueError("Desired level does not exist.")
+            level_to_fetch = self.LEVEL_MAP[level_to_fetch]
+
```
jeffreyoldham, line 103: Can we eliminate these two lines, directly using the value?

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -44,17 +45,19 @@ def __init__(self):
...
         for geoid, coords in self.downloader.iter_subareas():
-
             # This an admitedly hacky way of getting rid of bogus geographies
             # that are, for some mysterious reason, included as part of the US
             # states query (with geoIds like geoId/7000).
             # TODO(jeffreyoldham): (potentially) find a cleaner way to do this.
-            if int(geoid.split('/')[1]) > 100:
-                continue
+            # if int(geoid.split('/')[1]) > 100:
+            #     continue
```
jeffreyoldham, line 60: Why are these commented out?

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -63,8 +66,11 @@ def generate_simple_geojsons(self, show=False, eps=None):
                      tempfile.TemporaryFile(mode='r+') as f2:
```
jeffreyoldham, line 66: Is this line indented correctly?

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -63,8 +66,11 @@ def generate_simple_geojsons(self, show=False, eps=None):
...
 
                     geojson.dump(coords, f1)
-                    simple = self.simplifier.simplify(coords, epsilon=self.eps)
-                    geojson.dump(simple, f2)
+                    s, o_sz, c_sz = self.simplifier.simplify(coords,
+                                                             epsilon=self.eps)
+                    original_pts += o_sz
+                    compressed_pts += c_sz
+                    geojson.dump(s, f2)  # Save simple geojson in tempfile
```
jeffreyoldham, line 73:
```suggestion
                    geojson.dump(s, f2)  # Save simple geojson in tempfile.
```

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -73,10 +79,12 @@ def generate_simple_geojsons(self, show=False, eps=None):
...
                                           gpd.read_file(f2),
                                           show=show)
-                    self.simple_geojsons[geoid] = simple
+                    self.simple_geojsons[geoid] = s
             except AssertionError:
                 logging.error("Simplifier failure on GeoJSON below:\n", coords)
-        # plt.show()
+        logging.info(f"Total original points = {original_pts}\n"
+                     f"Compressed points = {compressed_pts}\n")
+        plt.show()
```
jeffreyoldham, line 87: Do we always want to do this?

scripts/us_census/geojsons_low_res/generate_mcf.py
```diff
@@ -34,6 +34,7 @@
 
 class McfGenerator:
     PLACE_TO_IMPORT = 'country/USA'
+    LEVELS_DOWN = 2  # counties
```
jeffreyoldham, line 37: What about `LEVELS_DOWN = 1`?

scripts/us_census/geojsons_low_res/simplify.py
```diff
@@ -92,7 +92,7 @@ def simplify(self, input=None, epsilon=0.01):
         logging.info(f"Original number of points = {original_size}.")
         logging.info(f"Simplified number of points = {simplified_size}.")
 
-        return geojson
+        return geojson, original_size, simplified_size
```
jeffreyoldham, line 95: Please update the `Returns` comment.

scripts/us_census/geojsons_low_res/download.py
fpernice-google: Done.

scripts/us_census/geojsons_low_res/download.py
fpernice-google: The need for this for loop is that if the user specifies e.g. level=2, we have to go two levels deep by composing the map LEVEL_MAP twice. Thus we go from `Country` -> `AdministrativeArea1` -> `AdministrativeArea2`, i.e. counties.

scripts/us_census/geojsons_low_res/download.py
fpernice-google: See comment above.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: This value needs to be changed when importing counties (level=2) vs states (level=1). I could make it a command-line argument, but importing states vs counties seems like two different enough operations that they would warrant an explicit change in the code. If you think a command-line argument would be helpful, I can add it though.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: This filter removed undesired state geoIds. Since in this case we're importing counties, this constraint doesn't make sense (all county geoIds are above 100).

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: I think so. My linter complains if I move it away from where it is.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: Done.

scripts/us_census/geojsons_low_res/generate_mcf.py
fpernice-google: In the case of logging, I am always doing it since one needs the flag `--alsologtostderr` to display logs anyways. About the `plt.show()` command, this is the command that shows a final plot where one can see all simplified geographies together. Currently, the tool always does this, even if the `show` flag is turned off. Turning the `show` flag on also shows individual comparisons for each simplified geography.

scripts/us_census/geojsons_low_res/simplify.py
fpernice-google: Done.

scripts/us_census/geojsons_low_res/generate_mcf.py
jeffreyoldham: Let's make it a command-line option.  We want people to be able to use the code without having to modify it.

scripts/us_census/geojsons_low_res/generate_mcf.py
jeffreyoldham: Let's protect with it the appropriate conditional for states rather than commenting it out.

fpernice-google:
Made final comment changes by @jeffreyoldham .

scripts/us_census/geojsons_low_res/download.py
```diff
@@ -70,7 +70,7 @@ def __init__(self):
         dc.set_api_key('dev')
```
tjann, line 70: Could probably rename it to describe what it's for (getting the **desired** next sub admin area). Feel free to postpone this to another PR.

scripts/us_census/geojsons_low_res/download.py
```diff
@@ -70,7 +70,7 @@ def __init__(self):
...
         self.geojsons = None
 
-    def download_data(self, place='country/USA'):
+    def download_data(self, place='country/USA', level=1):
```
tjann, line 73: Github won't let me comment above this line. Please update "# TODO(fpernice-google): Support downloading more than just US states." from line 25.

scripts/us_census/geojsons_low_res/download.py
tjann: I wonder how this works for weird cases like 5 counties are contained in NYC?

tjann:
For future GitHub work, would be good to keep an eye out for these build failures and figure out what is wrong: https://screenshot.googleplex.com/A8BzKVxEr75Qyit (in this case, yapf style again)

tjann (dismissed):
The counties are labeled as typeOf: dcs:State. Can you make the change to County?

fpernice-google (commented):
Addressed comments by @tjann.

jeffreyoldham (approved):
Good.  Thanks.

jeffreyoldham (approved):
Thanks.

PR #290, opened 2020-09-02
Rerun OECD with latest placeid2dcid mappings.

tjann:

PR #292, opened 2020-09-04
OECD: Add one more stat to break down resolved DCIDs by namespace.

tjann:
Cannot break down errors by namespace since error=no dcid=no namespace.


scripts/oecd/regional_demography/stats.txt
```diff
@@ -1,57 +1,70 @@
...
+wikidataId/Q160734      2
+geoNamesId/2064513      2
+wikidataId/Q192232      2
+geoNamesId/1629652      2
+iso/RO-NT               2
 Name: dcid, dtype: int64
 
-3. Number of OECD places mapped to each DCID, grouped by namespace (only those with N to 1 issues):
+3. Number of OECD places mapped to each DCID, grouped by namespace (only those
+   with N to 1 issues):
```
pradh, line 67: Somewhere could note that these N to 1 cases bad and generally should be filtered downstream to avoid inconsistent stats.

scripts/oecd/regional_demography/stats.txt
```diff
@@ -184,3 +197,13 @@ wikidataId  wikidataId/Q10686       2
...
+
+4. Number of resolved DCIDs by namespace (ignoring N to 1 issues):
+wikidataId    832
+nuts          748
+nan           371
+geoNamesId    274
+iso           266
+country        76
+geoId          53
+Name: namespace, dtype: int64
```
pradh, line 209: move this up? 1. and this are basically what we look at. And downstream logic should need to resolve the N to 1 issues. (Do OECD scripts do this?)

pradh (approved):

scripts/oecd/regional_demography/stats.txt
tjann: Noted in https://github.com/datacommonsorg/data/pull/292/files#diff-e8ecb27ad6899aad2e5dd75679de28d0R9-R15

scripts/oecd/regional_demography/stats.txt
tjann: Moved. Not sure if you meant for me to add `1. and this are basically what we look at.` I interpreted as "move this up? this is basically what we look at. " If what I changed is OK, feel free to merge this PR, then I will sync and update #290

pradh: This is ok. I meant #1 and <this-one> are the two stats to diff for validation. This is good, thanks.

PR #295, opened 2020-09-06
BLS CPI: schema for importing series for specific expenditure types

intrepiditee:
The existing imports only have the series for all expenditure types combined. This PR adds the schema needed for importing series for specific expenditure types.

Thanks for reviewing.


schema/bls.mcf
```diff
@@ -15337,18 +15337,86 @@ rangeIncludes: schema:Number
...
+typeOf: dcs:UnitOfMeasure
+description: "The reference base is December 1982 equals 100."
+descriptionUrl: "https://www.bls.gov/cpi/technical-notes/home.htm"
+
+Node: dcid:IndexPointBasePeriod1967Equals100
+typeOf: dcs:UnitOfMeasure
+description: "The reference base is 1967 equals 100."
+descriptionUrl: "https://www.bls.gov/cpi/technical-notes/home.htm"
+
+Node: dcid:InexPointBasePeriod1982To1984Equals100
```
tjann, line 15365: s/Inex/Index/g and check if the diffs look better (now it shows you removed a node when you're just trying to add others.

schema/bls.mcf
```diff
@@ -15337,18 +15337,86 @@ rangeIncludes: schema:Number
...
+description: "The reference base is December 2007 equals 100."
+descriptionUrl: "https://www.bls.gov/cpi/technical-notes/home.htm"
+
+Node: dcid:IndexPointBasePeriodDecember2009Equals100
+typeOf: dcs:UnitOfMeasure
+description: "The reference base is December 2009 equals 100."
+descriptionUrl: "https://www.bls.gov/cpi/technical-notes/home.htm"
+
 Node: dcid:IndexPointBasePeriodDecember1999Equals100
-name: "IndexPointBasePeriodDecember1999Equals100"
```
tjann, line 15347: Do you think there exists a human readable name that might concisely hint at what this means without looking at the description? If so we might be able to use it in charts, etc.

schema/bls.mcf
intrepiditee: fixed

schema/bls.mcf
intrepiditee: Maybe ReferenceBaseDecember1999Equals100 and get rid of the description?

tjann: Hmm.. slightly prefer current over that proposal. WBU?

intrepiditee: Yeah I guess I don't really see the problem with the current name. If we want the name to be short, at least "December1999Equals100" needs to be in the name.

schema/dcschema_enum_instances.mcf
```diff
@@ -123,11 +123,1601 @@ description: "Category of race or ethnicity group used by Opportunity Insights t
 typeOf: dcs:OI_RaceCodeEnum
 
 ###---- begin BEA enum instances ----###
-Node: dcid:ConsumerGoodsAndServices
```
tjann, line 126: Did you mean to delete this node?

schema/bls.mcf
tjann: Just to be clear--Currently there is no name (just dcid).

tjann (approved):
LG, just small question on intentionality of deleting ConsumerGoodsAndServices

schema/dcschema_enum_instances.mcf
intrepiditee: Yes because it is not used, or do we still want to keep it?

schema/bls.mcf
intrepiditee: I meant in the identifier. What are the advantages of having a name? I thought the dcid would still be displayed so we really just needed a descriptive dcid

tjann: Names are useful when the dcids are not human readable. This is a borderline case, but thanks for reminding me--the enums in this PR should get names.

schema/dcschema_enum_instances.mcf
```diff
@@ -123,11 +123,1601 @@ description: "Category of race or ethnicity group used by Opportunity Insights t
...
-name: "ConsumerGoodsAndServices"
+Node: dcid:BLS_SEFV04
 typeOf: dcs:EconomicProductEnum
-description: "Consumer goods are usually tangible final commodities purchased by the average consumer. Consumer services are intangible activities that are typically produced by other people and consumed by the consumer simultaneously."
-descriptionUrl: "https://en.wikipedia.org/wiki/Goods_and_services"
+description: "Food from vending machines and mobile vendors"
+
+Node: dcid:BLS_SAF11
+typeOf: dcs:EconomicProductEnum
+description: "Food at home"
```
tjann, line 132: Please add a name to these enums, similar to http://browser.datacommons.org/browser/NAICS/92

These are helpful in StatVar and StatPop displays: https://datacommons.org/browser/Count_Worker_NAICSAccommodationFoodServices

intrepiditee: Ok, I will turn the descriptions to names

intrepiditee: done. I moved them to the bottom of the file

schema/dcschema_enum_instances.mcf
```diff
@@ -1376,3 +1371,1599 @@ typeOf: dcs:ICD10Section
 Node: dcid:Overweight
 name: "Overweight"
 typeOf: dcs:HealthBehaviorEnum
+
+Node: dcid:BLS_SS47021
```
tjann, line 1375: 1. Is there a specific name for this set of product enums? Similar to how BLS publishes NAICS which is for industries. If there is, we could replace BLS with the more specific name.
2. Can you actually put this into a separate file? There's enough I think to warrant one.

Will merge this PR soon unless there are some concerns that I am not aware of/forgotten.

@intrepiditee

intrepiditee: 1. BLS just refers to them as items in this file https://download.bls.gov/pub/time.series/cu/cu.item.
2. Will create a new file

tjann: Ok. Can we move it to a new file and use BLSItem/SS47021 etc? LMK if you want to do it here, or in a follow up PR @intrepiditee

intrepiditee: done

tjann (approved):

PR #300, opened 2020-09-11
WDI: Add Earth place and remove indicators that overlap with Hanlu's WDI import

tjann:


pradh (approved):
Total pop for rest of the places come from google3 script right?

tjann:
> Total pop for rest of the places come from google3 script right?

Yeah, but I did it in a sketchy way for this round. I removed all geos except Earth when I did the import.

This adds to the sketchiness here that not all indicators are currently served in prod. (P1 still under review).

After our offline discuss, hold off on this PR--we may remove Total Pop indicator from here.

PR #309, opened 2020-09-23
[DPC COVID] Avoiding appending to existing files

intrepiditee:
Thanks for reviewing.


scripts/pcm-dpc/covid-19/preprocess.py
```diff
@@ -32,7 +38,7 @@ def generate_tmcf(self):
                     'observationDate: C:pcm-dpc->Date\n'
                     'value: C:pcm-dpc->{SVname}\n\n')
         idx = 1
-        with open(self.name + '.tmcf', 'a') as f_out:
+        with open(out, 'a') as f_out:
```
pradh, line 41: curious, why not just replace 'a' with 'w' ?

pradh (approved):
Can we skip the diffs to the CSV files?

intrepiditee:
> 
> 
> Can we skip the diffs to the CSV files?

dropped

scripts/pcm-dpc/covid-19/preprocess.py
intrepiditee: `geo_template` first writes to the output file, and then `generate_tmcf` appends to it

scripts/pcm-dpc/covid-19/preprocess.py
```diff
@@ -24,6 +26,10 @@ def preprocess(self):
 
     def generate_tmcf(self):
         """Generate the template mcf."""
+        out = self.name + '.tmcf'
+        if os.path.exists(out):
+            os.remove(out)
+
```
jeffreyoldham, line 32: Perhaps it would be better for the class to have the invariant that it only produces the tMCF file once, relying on its own logic, not on the filesystem.  This will make it clearer that the file contents are the same for every run.

E.g., it is not clear that self.geo_template() always produces the same output.

scripts/pcm-dpc/covid-19/preprocess.py
jeffreyoldham: Perhaps geo_template could be passed the file handle to which to write.

jeffreyoldham (commented):
Thanks for the quick turnaround.

scripts/pcm-dpc/covid-19/preprocess.py
intrepiditee: I did what you suggested below

PR #313, opened 2020-09-25
Fix eurostat edu enrollment StatVar MCF typo by removing the file

tjann:
There was a typo in the StatVar MCF in GitHub but not in g3. I removed the StatVar MCF here in favor of the ones in g3. Here are some screenshots for validation (I resolved without extra StatVar, just relying on g3):

image

image


scripts/world_bank/output/WorldBank_StatisticalVariables.mcf
```diff
@@ -0,0 +1,557 @@
+
+Node: dcid:WorldBank/SE_TER_CUAT_BA_FE_ZS
```
pradh, line 2: Is this change intentional?

tjann: No, reverted.

pradh (approved):

PR #320, opened 2020-10-08
Add US State, Intl Cities, and countries cohorts.

tjann:


beets (approved):
thanks!

scripts/datacommons/place_cohorts/intl/place_page_compare_countries.mcf
```diff
@@ -0,0 +1,66 @@
+
+Node: country/USA
+dcid: "country/USA"
+typeOf: schema:Place
```
shifucun, line 4: why not typeOf Country?

shifucun (approved):

scripts/datacommons/place_cohorts/intl/place_page_compare_countries.mcf
tjann: The code is general, to enforce best practice of e.g. not causing SF being labeled as City and County. We were going to/already added constraints to only allow typeOf from trusted provenances, but this is still better practice if the point is not to add new typeOf.

PR #335, opened 2020-11-19
Make india property countryAlpha2Code instead of isoCode

tjann:
The property `isoCode` does not exist for India (there are 2 representations of isoCodes for countries, ISO alpha-2 and ISO alpha-3), instead, it is called countryAlpha2Code. https://www.iso.org/iso-3166-country-codes.html

### DIFF
--- a/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.tmcf
+++ b/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.tmcf
@@ -8,4 +8,4 @@ value: C:COVID19_tests_india->CumulativeCount_MedicalTest_ConditionCOVID_19
 
 Node: E:COVID19_tests_india->E1
 typeOf: schema:Country
-isoCode: C:COVID19_tests_india->isoCode
+countryAlpha2Code: C:COVID19_tests_india->isoCode


tjann:
Before manifesting the COVID India Tests data, we will also need to modify the process_csv script to remove the duplicate stats for each date see https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.csv

I think the source data reports an AM and a PM stat. We could either take the later stat, or make the timestamp more accurate, such as 2020-11-19T00:00:00+05:30 and 2020-11-19T12:00:00+05:30. Thoughts?

pradh:
> Before manifesting the COVID India Tests data, we will also need to modify the process_csv script to remove the duplicate stats for each date see https://github.com/datacommonsorg/data/blob/master/scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.csv
> 
> I think the source data reports an AM and a PM stat. We could either take the later stat, or make the timestamp more accurate, such as 2020-11-19T00:00:00+05:30 and 2020-11-19T12:00:00+05:30. Thoughts?

+1 for taking the later data and keeping date granularity.

scripts/covid19_india/medical_tests_in_data/COVID19_tests_india.tmcf
```diff
@@ -8,4 +8,4 @@ value: C:COVID19_tests_india->CumulativeCount_MedicalTest_ConditionCOVID_19
 
 Node: E:COVID19_tests_india->E1
 typeOf: schema:Country
-isoCode: C:COVID19_tests_india->isoCode
\ No newline at end of file
+countryAlpha2Code: C:COVID19_tests_india->isoCode
```
pradh, line 11: Might need to add it to http://google3/datacommons/import/mcf_vocab.h?l=234-236&rcl=333437561 too?

pradh (approved):

tjann:
Per discussion offline, we will add isoCode property to countries, by duplicating the alpha2 code.

PR #337, opened 2020-11-26
WDI P1 from Ian

tjann:
Check in artifacts from importing the P1 set of variables


tjann:
@IanCostello FYI the resolver fails on populations like

Node: gen_pop_country/MLT_dcid:WorldBank/AG_LND_PRCP_MM
location: dcid:country/MLT
typeOf: dcid:StatisticalPopulation

when there is no populationType, which is required for StatPops.

We may need to fiddle with the CSV schema approach for direct place obs.

pradh:
Could you add a description on the changes?

pradh (approved):

PR #367, opened 2021-01-18
India Census: Primary Census Abstract Data for Houseless Population

thejeshgn:
1. India Census: Primary Census Abstract Data for Houseless Population 
2. Handled missing case of literacyStatus
3. Added more test cases for base class
4. Regenerated files


pradh (approved):
Nice!

PR #373, opened 2021-01-19
clean covid related imports generated CSV & TMCF

hanlu09205:
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/data/373)


scripts/covid_tracking_project/historic_state_data/COVIDTracking_States_StatisticalVariables.mcf
```diff
@@ -1,95 +0,0 @@
-Node: dcid:CumulativeCount_MedicalTest_COVID_19
```
pradh, line 1: Doesn't look like this is generated. I think this is hand-curated. Let's keep this?

pradh (approved):

scripts/covid_tracking_project/historic_state_data/COVIDTracking_States_StatisticalVariables.mcf
hanlu09205: I also see this defined in google3/medical/records/covid19/export/schema/Google_Health_COVID19_US_StatisticalVariables.mcf. 

Wonder whether the MCF is needed.

pradh: > I also see this defined in google3/medical/records/covid19/export/schema/Google_Health_COVID19_US_StatisticalVariables.mcf.
> 
> Wonder whether the MCF is needed.

The MCF should probably be subsumed by curated-stat-vars we have checked-in, but I think its better to confirm that is the case before we remove it.

It seemed like this PR is for deleting auto-gen files, and I was pointing out StatVar MCF isn't.

PR #384, opened 2021-01-22
Census India: Use placeOfResidenceClassification - Fix

thejeshgn:
- Use property `placeOfResidenceClassification`
- Regenrate census MCFs


pradh (approved):
Thanks for the fix!

PR #385, opened 2021-01-22
RBI BPL: Use property placeOfResidenceClassification

thejeshgn:
* Use property placeOfResidenceClassification
*  Similar to : https://github.com/datacommonsorg/data/pull/384


pradh (approved):
Thanks!

PR #397, opened 2021-02-10
Fix FBI city TMCF

hanlu09205:
Verified that for one pop_dcid, there is only one PopObsValidationResult. (Previously, there are two, with different obs_mmethod & obs_period

AapYuJFXrZBPcBE

Validate mix prod+new dashboard result is same as validate new. 
r7jxAqN5EnJG72C (1)

There is no duplicateobsval. 
(City is different from State, where legacy rape column number not included in the revised rape column. So there is no change to rape column for city stats.

### DIFF
--- a/scripts/fbi/crime/preprocess.py
+++ b/scripts/fbi/crime/preprocess.py
@@ -77,9 +77,9 @@ TEMPLATE_MCF_TEMPLATE = """
 Node: E:FBI_Crime->E{index}
 typeOf: dcs:StatVarObservation
 variableMeasured: dcs:{stat_var}
-measurementMethod: dcs:FBI_Crime
 observationAbout: C:FBI_Crime->GeoId
 observationDate: C:FBI_Crime->Year
+observationPeriod: "P1Y"
 value: C:FBI_Crime->{stat_var}
 """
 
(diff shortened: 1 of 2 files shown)


scripts/fbi/crime/preprocess.py
```diff
@@ -258,8 +258,6 @@ def _clean_crime_file(f_input, f_output):
         if field[_STATE_INDEX]:
             # Remove numeric values from state names (comes from footnotes)
             state = _remove_digits(field[_STATE_INDEX])
-            if year == "2016":
```
pradh, line 261: Why this diff?

scripts/fbi/crime/preprocess.py
```diff
@@ -77,9 +77,9 @@
 Node: E:FBI_Crime->E{index}
 typeOf: dcs:StatVarObservation
 variableMeasured: dcs:{stat_var}
-measurementMethod: dcs:FBI_Crime
 observationAbout: C:FBI_Crime->GeoId
 observationDate: C:FBI_Crime->Year
+observationPeriod: "P1Y"
```
pradh, line 82: For the record, the current MCF template is documented here: https://g3doc.corp.google.com/datacommons/mcf/crime/README.md?cl=355673889

pradh (approved):

scripts/fbi/crime/preprocess.py
hanlu09205: It was missed in previous PR during merge. 
Since 2016 table was updated with the right one, this is no longer needed.

PR #406, opened 2021-03-19
Update README.sh and fix python formatting

beets:
Update instructions for first time users - install git lfs, and setup python3
Ran yapf to fix all formatting

Might consider updating the cloudbuild configs to call a local script like we have in the website repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 __pycache__/
 *.DS_Store
 .vscode/settings.json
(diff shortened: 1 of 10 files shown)


pradh:
> Might consider updating the cloudbuild configs to call a local script like we have in the website repo

I can take a look...

pradh (approved):
Thanks for the change.

PR #410, opened 2021-03-23
Add an option to run place-name-resolver without GCS access

pradh:
Add a --generate_place_id option to produce placeID results instead of DCID.

### DIFF
--- a/tools/place_name_resolver/README.md
+++ b/tools/place_name_resolver/README.md
@@ -12,10 +12,20 @@ includes local-id references to contained-in places.
 
 For sample input/output CSVs, see the `.csv` files in `place_name_resolver/testdata` directory.
 
+NOTE: If the `--generate_place_id` is set, then in place of DCIDs, Maps placeIDs
+are returned. This is useful when you cannot access the GCS bucket storing the
+placeID to DCID map.
+
 ## Usage
 
 ```
 go run resolver.go --in_csv_path=testdata/input_basic.csv --out_csv_path=/tmp/output_basic.csv --maps_api_key=<YOUR_KEY_HERE>
+```
+
+For generating place ID:
+
+```
+go run resolver.go --in_csv_path=testdata/input_basic.csv --out_csv_path=/tmp/output_basic.csv --generate_place_id --maps_api_key=<YOUR_KEY_HERE>
 ```
 
 ## Testing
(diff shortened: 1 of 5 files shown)


tools/place_name_resolver/resolver.go
```diff
@@ -263,16 +268,19 @@ func writeOutput(outCsvPath string, tinfo *tableInfo) error {
...
-func resolvePlacesByName(inCsvPath, outCsvPath string, p2d PlaceId2Dcid, mapCli MapsClient) error {
+func resolvePlacesByName(inCsvPath, outCsvPath string, generatePlaceID bool, p2d PlaceId2Dcid, mapCli MapsClient) error {
 	tinfo, err := buildTableInfo(inCsvPath)
 	if err != nil {
 		return err
 	}
-	placeIdToDcid, err := loadPlaceIdToDcidMap(p2d)
-	if err != nil {
-		return err
+	placeId2Dcid := &map[string]string{}
```
shifucun, line 276: map is "pointer" like, can just do "map[string]string{}"

tools/place_name_resolver/README.md
```diff
@@ -12,12 +12,22 @@ includes local-id references to contained-in places.
 
 For sample input/output CSVs, see the `.csv` files in `place_name_resolver/testdata` directory.
 
+NOTE: If the `--generate_place_id` is set, then in place of DCIDs, Maps placeIDs
```
shifucun, line 15: i don't have too much context for this use, does it mean user could get the place id in batch (would that be a concern..)?

pradh: This is no different than accessing the public Maps API which return place IDs, which is what the tools does, right?  The user of the tool anyway needs to provide the Maps API Key.

tools/place_name_resolver/resolver.go
pradh: So its okay fro loadPlaceIdToDcidMap() to return a map, rather than a map pointer?  Wei had in the past suggested passing/returning pointers...

pradh (commented):
Thanks for the review!

tools/place_name_resolver/README.md
shifucun: Then that's okay.

tools/place_name_resolver/resolver.go
shifucun: Map itself is okay: https://stackoverflow.com/questions/40680981/are-maps-passed-by-value-or-by-reference-in-go

shifucun (approved):

tools/place_name_resolver/resolver.go
pradh: Thanks, changed all pointers to refs.

shifucun (approved):

PR #418, opened 2021-04-28
EIA common scraper with initial support for Electricity dataset

pradh:
This is an early half-done draft of EIA processing with partial schema support.

Quick sanity check:
https://datacommonsorg.github.io/tools/mcf-browser/#&file=https://raw.githubusercontent.com/pradh/data/eia_elec/scripts/us_eia/opendata/process/test_data/elec.tmcf&file=https://raw.githubusercontent.com/pradh/data/eia_elec/scripts/us_eia/opendata/process/test_data/elec.csv


scripts/us_eia/opendata/process/common.py
```diff
@@ -0,0 +1,167 @@
...
+        stats['info_lines_processed'] += 1
+        if stats['info_lines_processed'] % 100000 == 99999:
+          _print_stats(stats)
+
+        data = json.loads(line)
+
+        # Preliminary checks
+        series_id = data.get('series_id', None)
+        if not series_id:
+          stats['error_missing_series'] += 1
```
beets: these would be categories (there should only be these two types).. we might want to rename the counter?

scripts/us_eia/opendata/process/common.py
```diff
@@ -0,0 +1,167 @@
...
+      'typeOf: dcs:StatisticalVariable',
+      'populationType: schema:Thing',
+      f"measuredProperty: {raw_sv_id}",
+      'statType: dcs:measuredValue',
+  ])
+
+
+def process(in_json, out_csv, out_sv_mcf, out_tmcf, extract_place_statvar_fn,
+            generate_statvar_schema_fn):
+  """
```
beets: maybe describe the requirements for the lambda functions (even with [typing](https://stackoverflow.com/questions/33833881/is-it-possible-to-type-hint-a-lambda-function) it might not be too clear)

scripts/us_eia/opendata/process/common.py
```diff
@@ -28,6 +28,44 @@
...
+        return yr + '-' + _QUARTER_MAP[m_or_q]
+    else:
+      # Monthly
+      return yr + '-' + m_or_q
+
+  return None
+
+
+def _eia_dcid(raw_sv):
+  return 'dcid:eia/' + raw_sv.lower()
```
beets: oh this is nice

beets (approved):
this is cool!

scripts/us_eia/opendata/process/common.py
pradh: Ah I see. renamed to info.

scripts/us_eia/opendata/process/common.py
pradh: Added more details of args / return value for the lambda functions

pradh (commented):
Thanks for the review!

PR #421, opened 2021-05-01
EIA: Add preliminary schema-less support for Petroleum and Natural Gas datasets

pradh:


scripts/us_eia/opendata/process/test_data/ng.mcf
```diff
@@ -0,0 +1,17 @@
+Node: dcid:eia/ng.n3035_4.m
```
pradh, line 1: Do we want to keep this in its original capital case?

scripts/us_eia/opendata/process/common.py
```diff
@@ -53,27 +54,41 @@ def _parse_date(d):
...
+
     return None
 
 
 def _eia_dcid(raw_sv):
     return 'dcid:eia/' + raw_sv.lower()
 
 
+def _enumify(in_str):
+    return in_str.title().replace(' ', '')
```
beets, line 72: do we need to ensure camel case as well?

scripts/us_eia/opendata/process/ng.py
```diff
@@ -0,0 +1,66 @@
+"""EIA Natural Gas Dataset specific functions."""
+
+import re
+
+# List of series in Pattern #1 that include non-US countries. Unfotunately this
```
beets, line 5:
```suggestion
# List of series in Pattern #1 that include non-US countries. Unfortunately this
```

beets: also, could you describe the significance of the N/S prefix?

scripts/us_eia/opendata/process/test_data/ng.mcf
beets: yes, i think that would look more familiar and might be easier for exploration in the near term (e.g. if we want to look up a particular series in the browser)

beets (approved):
nice!

scripts/us_eia/opendata/process/common.py
pradh: title() approx does that?

https://stackoverflow.com/questions/3728655/titlecasing-a-string-with-exceptions

scripts/us_eia/opendata/process/test_data/ng.mcf
pradh: SG.  Stuck to original case

scripts/us_eia/opendata/process/ng.py
pradh: Done.

pradh (commented):
Thanks for the review!

scripts/us_eia/opendata/process/common.py
beets: oh, i missed that there was a call to "title" -- sg

PR #422, opened 2021-05-04
EIA: Fix date parsing

pradh:
Parse full date, which appears in PET dataset, correctly.


scripts/us_eia/opendata/process/common.py
```diff
@@ -38,6 +38,11 @@
 def _parse_date(d):
     """Given a date from EIA JSON convert to DC compatible date."""
 
+    try:
+      _ = int(d)
```
beets, line 42: could d.isnumeric() work here?

beets: out of curiosity, what kinds of non-numeric values did you come across? should we add some to the test?

beets (approved):
thx!

scripts/us_eia/opendata/process/common.py
pradh: I added it as an extra check, but there is actually no non-numeric value so far (ELEC, NG, PET).

Yes, isnumeric() is nicer!

pradh (commented):
Thanks!

scripts/us_eia/opendata/process/common.py
pradh: My comment about non-numeric is not true. There is 2019Q1 pattern. I'll add a test in upcoming PR.

PR #426, opened 2021-05-06
EIA: Add support for extracting StatVar names

pradh:
Extract StatVar names by removing the US place names in the string.

This seems to have pretty good coverage.  Missing ~10 stat-vars in INTL which do not have US country.  We can handle those by matching against other country names too, but skipping those for now.

Updated autopush SV hierarchy (Miscellaneous -> Variables):
- https://autopush.datacommons.org/browser/geoId/06
- https://autopush.datacommons.org/browser/country/USA


beets (approved):
Wow, looks really good! I saw these corner cases tho:

- Oxygenate Plant Production of Oxygenates (excluding Fuel Ethanol), Annual
- Renewable Fuels Plant and Oxygenate Plant Net Production of Other Oxygenates, Annual)
(maybe it's just Oxygenate that needs to be special cased, see https://autopush.datacommons.org/browser/country/USA?statVar=eia/PET.M_EPOOXO_YNP_MBBL.A)

pradh:
> Wow, looks really good! I saw these corner cases tho:
> 
> * Oxygenate Plant Production of Oxygenates (excluding Fuel Ethanol), Annual

This seems to be:  https://www.eia.gov/opendata/qb.php?sdid=PET.M_EPOOXXFE_YOP_NUS_MBBL.A

Interesting that this series is missing in BQ.  Wonder if its newer...

> * Renewable Fuels Plant and Oxygenate Plant Net Production of Other Oxygenates, Annual)
>   (maybe it's just Oxygenate that needs to be special cased, see https://autopush.datacommons.org/browser/country/USA?statVar=eia/PET.M_EPOOXO_YNP_MBBL.A)

Here too only the U.S. is stripped out, otherwise the sentence and case is as it is in source:
https://www.eia.gov/opendata/qb.php?category=432&sdid=PET.M_EPOOXO_YNP_NUS_MBBL.A

Hmm... so I don't follow what about Oxygenate need sspecial casing?

pradh:
Thanks for the review!

beets:
> Hmm... so I don't follow what about Oxygenate need sspecial casing?

My bad, goes to show how much more there is to learn about the space. It sounded like it should be "Oxygenating Plant", though [Oxygenates](https://www.api.org/oil-and-natural-gas/environment/clean-water/ground-water/oxygenates) is a Petroleum industry term, and [Oxygenating](https://www.gardeningknowhow.com/ornamental/water-plants/wgen/oxygenating-pond-plants.htm) is from gardening...

beets:
> > * Renewable Fuels Plant and Oxygenate Plant Net Production of Other Oxygenates, Annual)
> >   (maybe it's just Oxygenate that needs to be special cased, see https://autopush.datacommons.org/browser/country/USA?statVar=eia/PET.M_EPOOXO_YNP_MBBL.A)

For this as well, I couldn't tell the difference between the two charts (is there a disambiguating property we should show in the UI)? Also need to check why the observation nodes aren't showing loading..

pradh:
> > > * Renewable Fuels Plant and Oxygenate Plant Net Production of Other Oxygenates, Annual)
> > >   (maybe it's just Oxygenate that needs to be special cased, see https://autopush.datacommons.org/browser/country/USA?statVar=eia/PET.M_EPOOXO_YNP_MBBL.A)
> 
> For this as well, I couldn't tell the difference between the two charts (is there a disambiguating property we should show in the UI)? Also need to check why the observation nodes aren't showing loading..

This is very strange. They only differ by the provenance: https://screenshot.googleplex.com/3tgx3U2cpjfuNtT

eia/PET stuff should not show up from EIA_Electricity... I wonder if I messed up the import paths or something, I'll double check.

pradh:
I'll merge this PR.  Looking into dup chart separately...

PR #427, opened 2021-05-06
[EIA] Add processor for COAL dataset

beets:
Added some schema, but mostly schema-less. Of those, ignore many series which involved two places (exports, imports, and shipments).

Added schema here for review so that it makes sense with the code, will remove it before merging.


scripts/us_eia/opendata/process/common.py
```diff
@@ -83,8 +108,9 @@ def _enumify(in_str):
 
 def _print_counters(counters):
     print('\nSTATS:')
-    for k, v in counters.items():
-        print(f"\t{k} = {v}")
+    for k in sorted(counters):
+        # for k, v in counters.items():
```
pradh, line 112:
```suggestion
```

scripts/us_eia/opendata/process/test_data/COAL.csv
```diff
@@ -0,0 +1,2879 @@
+place,stat_var,date,value,unit,scaling_factor,eia_series_id
```
pradh, line 1: I wonder if we should curate the input COAL.txt a bit, to make it more readable/editable?

I have been trimming series to 2-3 data points, and picked 2 or 3 series that span patterns.

scripts/us_eia/opendata/process/schema.mcf
```diff
@@ -0,0 +1,63 @@
+# New units
+
+Node: dcid:BtuPerPound
+typeOf: dcs:EnergyUnitOfMeasure
+description: "Energy, in BTU's, per pound of fuel."
+name: "BtuPerPound"
+
+Node: dcid:ShortTon
```
pradh, line 8: Interesting!  I think in the Electricity dataset too I should use ShortTon instead of MetricTon.

From https://www.eia.gov/opendata/qb.php?category=446&sdid=ELEC.CONS_EG.PC-US-99.M its not clear whether its Metric or Short, because the glossary refers to both as Ton (https://www.eia.gov/energyexplained/units-and-calculators/), instead of Tonne (what wikipedia article suggests).

But matching up the Petroleum Coke numbers from https://www.eia.gov/totalenergy/data/monthly/pdf/sec7_9.pdf, it does seem to be ShortTon too... I'll make the change in a follow up PR.

scripts/us_eia/opendata/process/main.py
```diff
@@ -18,13 +32,15 @@
...
-    elif dataset == 'INTL':
+    if dataset == 'INTL':
         return intl.extract_place_statvar
-    elif dataset == 'PET':
+    if dataset == 'PET':
         return pet.extract_place_statvar
-    elif dataset == 'NG':
+    if dataset == 'NG':
         return ng.extract_place_statvar
     elif dataset == 'SEDS':
```
pradh, line 45:
```suggestion
    if dataset == 'SEDS':
```

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+    # Region could include 3-leter codes (e.g. MAT - Middle Atlantic)
+    m = re.match(r"^COAL\.([^._]+_?[^._]+)\.([A-Z]+)-([0-9]+)\.([AQM])$",
+                 series_id)
+    if m:
+        measure = m.group(1)
+        place = m.group(2)
+        code = m.group(3)
+        period = m.group(4)
+        sv_id = f'COAL.{measure}.{code}.{period}'
+        return (place, sv_id, True)
```
pradh, line 40: For context "is_us_place" exists because some datasets have 2 letter country codes which conflict with state code.  In this case, what all can place be?

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+        activity = m.group(1)
+        measure = m.group(2)
+        if activity in ['EXPORT', 'IMPORT']:
+            # Pattern #2
+            # TODO: model destination / source port as well
+            type = m.group(3)
+            place = m.group(4)
+            port = m.group(5)
+            period = m.group(6)
+            sv_id = f'COAL.{activity}_{measure}.{type}.{period}'
```
pradh, line 56: If we don't add port to the sv_id, could we end up with inconsistent series where for the same date and everything else the values are different ?

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+            return (place, sv_id, False)
+        elif activity == 'SHIPMENT':
+            # Pattern #3
+            source = m.group(3)
+            if source.isalpha():  # could include 3-letter region codes
+                destination_power_plant = m.group(4)
+                material = m.group(5)
+                period = m.group(6)
+                sv_id = f'COAL.SHIPMENT_{measure}.{material}.{period}'
```
pradh, line 65: Ditto with destination_power_plant

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+        'statType: dcs:measuredValue',
+    ],
+    'COST': [
+        'Average_Cost_Coal',
+        'populationType: dcs:Coal',
+        'measuredProperty: dcs:cost',
+        'statType: dcs:meanValue',
+    ],
+    'HEAT_CONTENT': [
+        'AverageQuality_HeatContent_ReceivedCoal',
```
pradh, line 122: Just "Average" here?

(I thought AverageQuality was used for ash content since it talks about quality of coal)

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+    # COAL.{Measure}.{ConsumingSector}.{Period}
+    m = re.match(r"^COAL\.([^._]+_?[^._]+)\.([0-9]+)\.([AQM])$", raw_sv)
+    if m:
+        measure = m.group(1)
+        consuming_sector = m.group(2)
+        period = m.group(3)
+    else:
+        counters['error_unparsable_raw_statvar'] += 1
+        return False
+    counters[f'measure-{measure}'] += 1
```
pradh, line 191: nice idea!

scripts/us_eia/opendata/process/coal.py
```diff
@@ -0,0 +1,236 @@
...
+            row['unit'] = f'dcid:{unit}'
+        else:
+            # Reset unit to empty to clear the raw unit value.
+            row['unit'] = ''
+        if sfactor:
+            row['scaling_factor'] = sfactor
+
+    if sv_id not in sv_map:
+        node = f'Node: dcid:{sv_id}'
+        sv_map[sv_id] = '\n'.join([node] + sv_pvs)
```
pradh, line 234: Ah... this reminds me that it might be worth refactoring the interface a bit so this happens in common.

pradh (commented):
Looks good!

A few questions around the places and raw-sv-id gen, otherwise nits.

pradh:
Thanks for adding license!!

scripts/us_eia/opendata/process/test_data/COAL.csv
beets: oh good point. i selected a 1% random sample of the file. it's helped to catch some other errors while i was adding patterns. i'll trim down some of the datapoints, and reduce the number of series.

scripts/us_eia/opendata/process/schema.mcf
beets: wow, that's a nice find

scripts/us_eia/opendata/process/coal.py
beets: yup, these are all US places. we'll only have intl places for the export/import ones which we'll also need to resolve somehow (since only either dest/src are intl)

scripts/us_eia/opendata/process/coal.py
beets: good pt - leaving these out for now

scripts/us_eia/opendata/process/coal.py
beets: thx!

pradh:
If I look at diffs from commits since last review, I'm not seeing much.

https://github.com/datacommonsorg/data/pull/427/files/c6166b28abdfa5bcabc00b4544c1f10e5731ba96..4e0c7809bf975f866c8bcbdf3048f52479d76ad6

Is it perhaps because of the merge conflicts? (Probably because of https://github.com/datacommonsorg/data/pull/428)

beets:
Validated and spot checked some of the numbers -- there really are some charts with big variance
https://www.eia.gov/opendata/qb.php?category=717362&sdid=COAL.HEAT_CONTENT.US-8.Q
https://www.eia.gov/opendata/qb.php?category=717362&sdid=COAL.HEAT_CONTENT.ID-8.Q

Also fixed a bug with a category that had a null name. It gets added as "null" which will at least be clickable (shows up here as well, though as empty string: https://www.eia.gov/opendata/qb.php?category=2710708)

scripts/us_eia/opendata/process/coal.py
```diff
@@ -39,38 +39,40 @@ def extract_place_statvar(series_id, counters):
...
-                sv_id = f'COAL.SHIPMENT_{measure}.{material}.{period}'
-                return (source, sv_id, True)
-            else:
-                # TODO: Handle remaining places - coal mines
-                counters[f'error_unknown_coal_mine SHIPMENT '] += 1
-                return (None, None, None)
-        else:
-            counters[f'unknown #2,3 activity ({activity})'] += 1
-            return (None, None, None)
+    # m = re.match(r"^COAL\.([A-Z]+)_([A-Z]+)\.([^-]+)-([^-]+)-([^.]+)\.([AQM])$",
```
pradh, line 45: nit: maybe delete this with a link to https://github.com/beets/data/blob/c6166b28abdfa5bcabc00b4544c1f10e5731ba96/scripts/us_eia/opendata/process/coal.py#L44-L73 for future reference

scripts/us_eia/opendata/process/test_data/COAL.csv
pradh: Thanks!

pradh (approved):
Thanks for the changes!  Sorry about the merge conflicts.

pradh:
> Validated and spot checked some of the numbers -- there really are some charts with big variance
> https://www.eia.gov/opendata/qb.php?category=717362&sdid=COAL.HEAT_CONTENT.US-8.Q
> https://www.eia.gov/opendata/qb.php?category=717362&sdid=COAL.HEAT_CONTENT.ID-8.Q

wow!

> Also fixed a bug with a category that had a null name. It gets added as "null" which will at least be clickable (shows up here as well, though as empty string: https://www.eia.gov/opendata/qb.php?category=2710708)

Thanks for the fix.

scripts/us_eia/opendata/process/coal.py
beets: I'd like to keep that, if it's ok? It took a while to figure out, and I sometimes go on cleaning sprees of idle branches so don't want to lose it. I hope we can uncomment them out soon..

beets:
thanks for the review!

PR #430, opened 2021-05-10
EIA open data fixes

beets:
- Fix error in COAL SV generation
- Split SVG into separate file; update tests
- Generate SVG for ELEC
- Fix extra dcid: prefix during SVG generation


pradh (approved):
Thanks for the dcid: catch!

beets:
np!

PR #431, opened 2021-05-11
EIA: add name and category SVG for schema-ful stat-vars

pradh:
Also:

+ Improve naming heuristics.
+ Fix tests (I'm very puzzled if the presubmit actually runs)


beets (approved):
thanks for the fix! and for fixing the tests..

btw, the file renames look mis mapped -- could you double check that git gets it right?

pradh (commented):
> thanks for the fix! and for fixing the tests..
> 
> btw, the file renames look mis mapped -- could you double check that git gets it right?

The mis-mapped ones are empty svg.mcf files, so its okay I think.

Thanks for the review!

PR #435, opened 2021-05-12
[EIA] Fix tests

beets:
- missing `scripts/us_eia/__init__.py` meant that the tests weren't discovered from run_tests.sh
- import modules in `__main__` files needs that awful line so that the relative modules can be picked up

This is an awfully brittle setup. We might need to rethink how the repo is setup...


scripts/us_eia/opendata/process/nuclear.py
```diff
@@ -0,0 +1,145 @@
+# Copyright 2021 Google LLC
```
pradh, line 1: This shouldn't be part of this PR?

scripts/us_eia/opendata/process/main.py
```diff
@@ -14,17 +14,18 @@
...
-import pet
-import seds
-import total
+# Allows the following module imports to work when running as a script
+# relative to scripts/
+sys.path.append(
+    os.path.dirname(
+        os.path.dirname(
+            os.path.dirname(os.path.dirname(os.path.abspath(__file__))))))
+from us_eia.opendata.process import coal, common, elec, intl, ng, nuclear, pet, seds, total
```
pradh, line 28: omg :)

pradh (approved):
> * missing `scripts/us_eia/__init__.py` meant that the tests weren't discovered from run_tests.sh
> * import modules in `__main__` files needs that awful line so that the relative modules can be picked up
> 
> This is an awfully brittle setup. We might need to rethink how the repo is setup...

Thanks for finding this out!

Until we find a better setup, is this worth documenting in repo's top-level README.md to ensure everyone adds presubmit runnable tests?

scripts/us_eia/opendata/process/nuclear.py
beets: merged in changes from other pr. i miss diffbase :-/

PR #438, opened 2021-05-14
Schema file path update

spiekos:
Update the schema filepath links in the data README.md files to their current locations in the github repository.


pradh (approved):
Thanks for the update!

PR #440, opened 2021-05-19
Fix COVID Tracking MedicalTest stat-var IDs

pradh:
Rename the Stat Var IDs to match what is in prod. Previously the same stat-var had a different ID for COVID Tracking.

TESTED=Regenerated the TMCF manually.


beets (approved):

pradh:
Thanks!

PR #443, opened 2021-05-30
Create README.md

spiekos:
Documentation for the ClinVar import


scripts/biomedical/ClinVar/README.md
```diff
@@ -0,0 +1,143 @@
+
+# Importing ontology dataset of molecular interaction from the European Bioinformatics Institute (EMBL-EBI)
```
pradh, line 2: Slightly puzzled by the README.md.

It seems like a combination of 
https://github.com/datacommonsorg/data/blob/master/scripts/biomedical/proteinInteractionEBI/README.md and https://github.com/datacommonsorg/data/blob/master/scripts/biomedical/dbSNP/README.md

That is intentional?

pradh (approved):

PR #448, opened 2021-06-25
[WorldBankCountryBoundaries] Add import notebook for country boundaries

beets:
Py Notebook that downloads and exports admin area 0 boundaries, in HD (original from source), and simplified (using [`geopandas.GeoSeries.simplified`](https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoSeries.simplify.html).

The generated MCF's are still rather large compared to similar files for US counties and states. At DP3, counties are 1.2MB, whereas this is 2.5MB. I generated MCF's for more levels in this notebook as well - we can test at DP3, and import the additional ones if necessary. See all the [generated MCF's in gcs](https://pantheon.corp.google.com/storage/browser/unresolved_mcf/template_mcf_imports/WorldBankBoundaries/mcfs/mcfs?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false).

To see the colab notebook directly, [use this link](https://colab.research.google.com/drive/16Sr7GjcCcmo9bnghDajujbOIlvdGg30s). It's a 25MB file exported, so I can reduce the output if that helps - lmk.


pradh:
Nice, and the viz stuff in the NB is very useful!

I wonder if you're thinking we use NBs for scripts as a model or just do that for this import?  Because tests are hard/painful with NB.

I'll take a closer look at the NB shortly.

beets:
I'm thinking just for this import - yes, testing is hard most of the time, though for this import it was really useful to visually inspect the images. But getting a notebook for data exploration and validation could be useful.

pradh:
> I'm thinking just for this import - yes, testing is hard most of the time, though for this import it was really useful to visually inspect the images. But getting a notebook for data exploration and validation could be useful.

In reviewing it (the first NB-based import), my feeling is that it is probably better to delineate code used for exploration/introspection from (production) code that actually generates the output.  The latter can be carefully reviewed, tested and re-run when necessary.

But I agree that seeing the exploration/validation definitely helps in capturing some wisdom involved in the import!

pradh (approved):

beets:
ptal @pradh - I moved the data processing code out to a script and added tests. Agree that this feels much more natural, and we should leave notebooks for exploration.

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
```diff
@@ -0,0 +1,202 @@
...
+    def build_import_codes(self, all_countries):
+        """Builds mapping of column code to country codes to import.
+
+        Only countries with DCID of the form county/{code} are included.
+
+        Returns:
+          A dict mapping dataframe column names to iterable of values in that column to include in the import. e.g.:
+          {'WB_A3': ['FRA', 'NDL',...],
+          'ISO_A3': ['UMI', 'COD', ...]}
+        """
```
pradh, line 111: super nit: for readability maybe one line space between docstring and this line? or a comment like "Call DC API to get ..."

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
```diff
@@ -0,0 +1,202 @@
...
+        }
+        ''',
+                                    select=None)
+
+        def is_dc_country(iso):
+            dcid = f'country/{iso}'
+            matches = filter(lambda x: x['?place'] == dcid, dc_all_countries)
+            return len(list(matches)) > 0
+
+        wb_idx = all_countries['WB_A3'].apply(is_dc_country)
```
pradh, line 125: maybe a comment somewhere as to what WB_A3 is?

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
```diff
@@ -0,0 +1,202 @@
...
+)
+flags.DEFINE_string(
+    'mcf_dir', '',
+    'Dir to output generated MCF files too. If blank, a temp folder will be used.'
+)
+
+DOWNLOAD_URI = 'https://development-data-hub-s3-public.s3.amazonaws.com/ddhfiles/779551/wb_boundaries_geojson_highres.zip'
+TMP_PATH = '/tmp/wb_boundaries_geojson_highres'
+GEO_JSON_PATH = 'WB_Boundaries_GeoJSON_highres/WB_countries_Admin0.geojson'
+ISO_A3_CODES_TO_IMPORT = ['COD', 'ROU', 'TLS', 'AND', 'IMN', 'UMI', 'USA']
```
pradh, line 57: maybe a comment on what this extra map is for?

pradh (commented):
Thanks for the move to src repo and tests! 

Somehow the diffs have a bunch of past changes (this happened for me too yest and I'm not fully sure why the PR doesn't get "refreshed" from master)...

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
beets: done!

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
beets: good point, added

scripts/world_bank/boundaries/country_boundaries_mcf_generator.py
beets: done, added that with the explanation of the extra map of country codes

beets:
thanks for the additional review!

PR #467, opened 2021-07-23
Auto-assign PRs using Blunderbuss

beets:
Use [Blunderbuss](https://github.com/googleapis/repo-automation-bots/tree/master/packages/blunderbuss) to auto-assign PR's to our core team, which is set up to round-robin assign PR's to team members. We can also add a label to auto-assign the backlog of open PR's.

### DIFF
--- a/.github/blunderbuss.yml
+++ b/.github/blunderbuss.yml
@@ -1,0 +1,2 @@
+assign_prs:
+  - datacommonsorg/core


.github/blunderbuss.yml
```diff
@@ -0,0 +1,2 @@
+assign_prs:
+  - datacommonsorg/core
```
pradh, line 2: Thank you for adding this!  Would it be worth having a different team, with more focussed reviewers? (e.g., Core has Rafik, Guha)

pradh (approved):

beets:
I've updated the setting on the team to route the reviews to a smaller
subset of the team, so Guha won't be getting reviews :)

beets:
Mis-typed - it does assign it to everyone on the team, and the setting I updated was for a different feature. Sent a followup #468

PR #480, opened 2021-08-13
Scripts for DeepSolar importing

katherinelamb:
Scripts for importing DeepSolar results and testing scripts


n-h-diaz (approved):
thanks! looks good :)

PR #451, opened 2021-07-01
NHM District level data: Maternal Health dataset

SenthamizhanV:
@pradh 
Commit contains district level data for Maternal health dataset (India)

PR #416, opened 2021-04-26
India NHM Data: Uploaded Child Health, Birth Control & General Stats datasets

SenthamizhanV:
@pradh Created a new PR for other three NHM datasets. Kindly take a look at it.

PR #414, opened 2021-04-13
India NHM Data: State-wise and District-wise data

SenthamizhanV:
@pradh @thejeshgn 

Added National Health Mission data for state-wise maternal data under `india_nhm` folder.

---
This change is [Reviewable](https://reviewable.io/reviews/datacommonsorg/data/414)

PR #482, opened 2021-08-18
NHM District Level Data: All datasets

SenthamizhanV:
@pradh 
PR containing all four datasets from NHM - Maternal Health, Birth Control, Child Health and General Statistics


pradh (approved):
This LG so far. Since its been sitting for a while, let me merge it in.

Does this obsolete https://github.com/datacommonsorg/data/pull/451 and https://github.com/datacommonsorg/data/pull/416? If so, can you please close the PRs.

What about the statewise data in https://github.com/datacommonsorg/data/pull/414?

SenthamizhanV:
This PR makes all the other three obsolete. Do we need state-wise data when we have district-wise? State-wise is just an aggregated dataset. I think we can close the state-wise PRs too.

pradh:
> This PR makes all the other three obsolete. Do we need state-wise data when we have district-wise? State-wise is just an aggregated dataset. I think we can close the state-wise PRs too.

Let us import the state-wise aggregate too.

PR #485, opened 2021-08-25
India: PLFS - July 2019 - Jun 2020  Updates

thejeshgn:
- July 2019 - Jun 2020 - Average daily wage earnings
- Handles both CSV and XLSX
- Uses standard function for resolving state codes
- Added more test cases


thejeshgn:
@pradh For some states the reported values are `0`. I think it's not zero but not available. Should we remove these from the final CSV?



dailywage-2021-08-25T20-17-39

PR #496, opened 2021-09-10
Refactor crosswalk functionality and integrate it into facility generation

pradh:
To get the download.py to run on py 3.6, needed a couple of tweaks. Notably, the default excel engine stopped supporting xlsx (https://stackoverflow.com/a/65266270).


beets (approved):
very cool, thanks for this refactor!

i've been running the download script with a repo specific venv and python 3.6, so i'm surprised there are still differences in behavior when we run it. we should dig into it when there's time :-/

pradh:
> very cool, thanks for this refactor!
> 
> i've been running the download script with a repo specific venv and python 3.6, so i'm surprised there are still differences in behavior when we run it. we should dig into it when there's time :-/

Thanks!   That is very puzzling... fwiw, my version is `Python 3.6.8`

I will submit this, and could you please try running the `download.py` on your local setup once to make sure I've not regressed something?

beets:
it still works for me, though i needed to add to requirements.txt - will send that in the next pr

PR #497, opened 2021-09-10
Add testing guidelines to README and consolidate EPA scripts under us_epa/

pradh:
Incorporate `__init__.py` in the older EPA scripts.


README.md
```diff
@@ -113,6 +113,41 @@ source .env/bin/activate
...
+framework](https://docs.python.org/3/library/unittest.html), and named with
+an `_test.py` suffix.
+
+A common test pattern is to drive your main processing function through some
+sample input files (e.g., with a few rows of the real csv/xls/etc.) and
+compare the produced output files (e.g., cleaned csv, mcf, tmcf) against
+expected ones.  An example test following this pattern is
+[here](https://github.com/datacommonsorg/data/blob/b868f558497783bba7a7f3ced9b918f1c0249290/scripts/us_epa/facility/process_facility_test.py).
+
+Please add an `__init__.py` file with your import scripts to include a
```
beets, line 128: we should mention that it's needed for the entire path from scripts/

beets (approved):
thanks for the documentation update and code cleanup!

README.md
pradh: Good point!

`scripts/` was actually missing `__init__.py`, which would've made the comment confusing. So I added that file in this PR.

pradh:
Thanks for the review!

PR #498, opened 2021-09-11
Add a bunch of __init__.py fixing tests along the way

pradh:


beets (approved):
nice, thanks for this! i was considering updates to use/support pytest, which won't require all the __init__.py's, but needed more work for modules that use absl.flags. this is great though!

pradh:
> nice, thanks for this! i was considering updates to use/support pytest, which won't require all the **init**.py's, but needed more work for modules that use absl.flags. this is great though!

absl.flags also forces all the flags across scripts to be unique, which is a pain. One more reason to move away from it.

I was going to follow this up with a presubmit check that new directories with tests have `__init__.py`... if the migration to pytest is imminent, I can hold off on that though.

Thanks for the review!

PR #500, opened 2021-09-14
[EIA Electricity] Scale values in place of setting scalingFactor incorrectly

pradh:
+ Change KilowattHour and MegawattHour to GigwattHour
+ Let percents be as is


pradh:
1. California StatVar summary

image

2. Spot-check

- original: 
image

- new: 
image

scripts/us_eia/opendata/process/elec.py
```diff
@@ -216,22 +216,22 @@ def extract_place_statvar(series_id, counters):
...
-    'CUSTOMERS': ('', ''),
-    'GEN': ('MegawattHour', '1000'),
-    'PRICE': ('USCentPerKilowattHour', ''),
-    'RECEIPTS': (_PLACEHOLDER_FUEL_UNIT, '1000'),
-    'RECEIPTS_BTU': ('Btu', '1000000000'),
-    'REV': ('USDollar', '1000000'),
-    'SALES': ('KilowattHour', '1000000'),
-    'STOCKS': (_PLACEHOLDER_FUEL_UNIT, '1000'),
-    'SULFUR_CONTENT': ('', '100'),
+    'ASH_CONTENT': ('', '100', 1),
```
beets, line 221: should we use https://datacommons.org/browser/Percent as a unit for this?

beets (approved):
thanks for the fix!

scripts/us_eia/opendata/process/elec.py
pradh: This I think used to be a legacy representation. So I'm slightly hesitant to use so that we're not adding one more slightly different way of representing percent.

For context, we have the following representations for percent:
1. unit = Percent
2. measuredProp = percent
3. measurementDenominator = X, scalingFactor = 100

We sort of deprecated (1) and (2) and moved to approach (3) in the more recent imports.

In (3), if instead of (or in addition to?) scalingFactor, we do unit = Percent, that will be another new way.

pradh:
Thanks for the review!

PR #501, opened 2021-09-14
Remove scaling factor from StatVarObs and update value

ajaits:
The multiplier for units is factored into the statVarObservation value and the scalingFactor property is removed.

Tested:
dc-import lint un/energy/test_data/un_energy_output.*


pradh (approved):
Thanks for the fix!

_get_scaled_value() is a good way of doing it!  I will use it in EIA too..

pradh (approved):

PR #502, opened 2021-09-15
[EIA] Produce DCID using epaGhgrpFacilityId and reference to Power Plant entities

pradh:
Multiple GHG Facility IDs can map to the same FRS ID or power-plant ID.  So we derive the DCIDs using the GHG facility ID, and attach the other properties as epaFrsId and PowerPlant reference (as partOf).


beets (approved):
thanks shanth!

pradh:
Thanks for the review!

PR #503, opened 2021-09-15
[EPA FLIGHT] fix sv names, update place dcid's

beets:
Incorporate changes discussed
- updated sv names / descriptions / properties
- use updated facility dcid's

There are only a handful of duplicate observations left, but those seem to be errors I can't trace (in the dataset).

report.json still returns a clean report.


pradh (approved):
Thanks for the fix!

beets:
removed duplicates (keeping first encountered) -- the sheets are processed in order. import now has no duplicates!

PR #505, opened 2021-09-15
Update cdc places script and test files

padma-g:
@pradh @spiekos 

Made an update to the script to put the cleaned csv file's columns in the right order


pradh (approved):
Thanks for the quick change!

pradh:
You will need to run yapf to fix format check failure

padma-g:
Fixed, thank you for reviewing!

PR #506, opened 2021-09-16
Improve lat/lng mapping for facilities

pradh:


beets (approved):
thanks for these fixes!

pradh:
Thanks for the review!

PR #508, opened 2021-09-18
Improve GHG Facilities geo relations by treating zip as source of truth

pradh:
The original bug (of facilities with lat/lng in CO showing up as contained in CA) happens when one FRS facility has multiple GHG facilities. In that case, the zip is of the GHG facility, but the lat/lng and county fips are of the FRS facility.

Example:
- [FRS Facility](https://ofmpub.epa.gov/frs_public2/fii_query_detail.disp_program_facility?p_registry_id=110043815057) in CO has GHG reporting facilities in 3 other states: WY, NE, AK
- Those facilities have the local zip, but the same CO county-fips and lat/lng:  [WY](https://enviro.epa.gov/cache/query/enviro3/efservice/PUB_DIM_FACILITY/facility_id/1006603/=/ROWS/0:0), [NE](https://enviro.epa.gov/cache/query/enviro3/efservice/PUB_DIM_FACILITY/facility_id/1006605/=/ROWS/0:0), [AK](https://enviro.epa.gov/cache/query/enviro3/efservice/PUB_DIM_FACILITY/facility_id/1006604/=/ROWS/0:0)

Thus this PR treats zip-code as the source-of-truth.

Compared to fixes in https://github.com/datacommonsorg/data/pull/506/files, this change maps ~172 more counties, but loses ~1000  more lat/lngs because we now only include them if we can establish containment in county. 

The 1000 lost lat/lngs roughly comes from ~600 where zips are not valid ZCTAs ("missing_zip_and_county") and from ~400 where we found a county from DC based on zip ("zipbased_county_wrong_latlng") and its lat/lng doesn't match.

```
Geo Resolution Stats: 
  given_county_wrong_latlng -> 66 - 1008370, 1002580, 1000586
  given_county_correct_latlng -> 9058 - 1001417, 1001552, 1002039
  zipbased_county_wrong_latlng -> 432 - 1002668, 1006603, 1001944
  zipbased_county_correct_latlng -> 339 - 1012147, 1001767, 1011963
  missing_zip_and_county -> 617 - 1002223, 1002506, 1002550
```


scripts/us_epa/facility/process_facility.py
```diff
@@ -112,36 +111,26 @@ def _get_naics(table, row):
...
+    """Returns counties that the zcta is associated with.
 
-def _is_nearby(polygon, point):
-    if polygon.geom_type == 'MultiPolygon':
-        for p in list(polygon):
-            if p.exterior.distance(point) <= _DISTANCE_THRESHOLD:
-                return True
-        return False
-    return polygon.exterior.distance(point) <= _DISTANCE_THRESHOLD
+       Returns: two candidate county lists corresponding to cip and geoOverlaps respectively.
```
beets, line 117:
```suggestion
       Returns: two candidate county lists corresponding to zip and geoOverlaps respectively.
```

scripts/us_epa/facility/process_facility.py
```diff
@@ -156,12 +145,67 @@ def _validate_geo(facility, dcid, lat, lng):
...
+    county_candidates = _get_county_candidates(zip)
+    if any(provided_county in l for l in county_candidates):
+        # Provided county is in the candidate list, use that.
+
+        if lat and lng and _validate_latlng(lat, lng, provided_county):
+            # Lat/lng is in the chosen county
+            _COUNTERS['given_county_correct_latlng'].append(facility_id)
+            return zip, provided_county, lat, lng
+
+        _COUNTERS['given_county_wrong_latlng'].append(facility_id)
```
beets, line 180: have you tried skipping this, and falling through if there is lat/lng defined (but the wrong county is entered?).

beets (approved):
thanks for this!

scripts/us_epa/facility/process_facility.py
pradh: This is the case where the county entered is likely right (because DC zip containment matches), but the lat/lng is slightly off.  I thought its better to use the county they report for stats, even if we skip the lat/lng (the alternative is to potentially pick a different zip-overlapping county with right lat/lng), and there are only 66 such cases.

But if the wrong county is entered, then the check in line 172 will not pass, so we will validate the candidate counties against lat/lng in the block below.

pradh (commented):
Thanks for the review!

scripts/us_epa/facility/process_facility.py
beets: gotcha - thanks sgtm!

PR #511, opened 2021-09-22
Constraint prefix edge case handling and style fixes in statvar_dcid_generator.py

Abilityguy:
cc: @ajaits , @pradh, @beets 
Handled a minor edge case in constraint prefix removal.
Refactored constraint property handling into a separate function.
Changed names to be more consistent with the file name (stat_var -> statvar) and other style fixes.


pradh (approved):

pradh:
Thanks for the fixes!

PR #521, opened 2021-10-04
ACS5yrSubjectTable: S0101

ajaits:
Import for Subject table: S0101.

Includes S0101_spec.json used to generate the statvars and 
test_data/ with sample input and expected outputs.


scripts/us_census/acs5yr/subject_tables/s0101/S0101_spec.json
```diff
@@ -0,0 +1,287 @@
...
+      "Age dependency ratio": "100",
+      "Child dependency ratio": "100",
+      "Old-age dependency ratio": "100",
+      "Sex ratio (males per 100 females)": "100"
+    }
+  },
+  "enumSpecializations": {
+  },
+  "inferredSpec": {},
+  "universePVs": [],
```
sharadshriram, line 96: Since `enumSpecializations`, `inferredSpec` and `universePVs` are empty, perhaps we can remove them?
I guess, the code to generate the column map handles the case where certain keys are missing by initialising them to empty dictionary (or list).

scripts/us_census/acs5yr/subject_tables/s0101/S0101_spec.json
```diff
@@ -0,0 +1,287 @@
...
+      "85 years and over",
+      "Under 18 years",
+      "Under 5 years"
+    ]
+  },
+  "ignoreColumns": [
+    "id",
+    "Geographic Area Name",
+    "Total MOE",
+#    "Margin of Error",
```
sharadshriram, line 135: Do you perhaps intend to comment `Margin of Error`? I do not think `#` is allowed in a valid JSON.

Also, since you use the token `Margin of Error` in [line 24](https://github.com/datacommonsorg/data/blob/afb56bec3771e65ca21117f66acb5420ae17646f/scripts/us_census/acs5yr/subject_tables/s0101/S0101_spec.json#L24), I think you can remove this line.

PR #529, opened 2021-10-13
S0701 import

rbhoot:
[Import Doc](https://docs.google.com/document/d/1LO6m9DZJpMz8uLWEYDuxBzF5Utki3U83hVjLaGVHAgE/edit?resourcekey=0-qSZUIaPfS2tIBvdhcm2Mow#heading=h.alsa48vfsv4c)


scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -27,7 +27,7 @@
 from .process import set_column_map
 
 # These directories are excluded from testing
-_EXCLUDE_DIRS = ['common', 's2201', '__pycache__']
+_EXCLUDE_DIRS = ['common', 's2201', '__pycache__', '~']
```
ajaits, line 30: is there a '~' directory in this path or does the os util to list dirs pull in home dir as well?

ajaits (commented):
Please add a short README describing the properties  covered in this table.

The unittest seems to have failed due to a floating point difff:
54.393999999998 vs 54.394

Can we restrict the number of decimal places in the script esp when computing values?

rbhoot:
> Please add a short README describing the properties covered in this table.
> 
> The unittest seems to have failed due to a floating point difff: 54.393999999998 vs 54.394
> 
> Can we restrict the number of decimal places in the script esp when computing values?

added README. I could only find 16054.394, if that is causing trouble unittest might need some change because test data has restricted number of decimal places. The unit test had passed when I checked.

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
rbhoot: no, macOS lists home directory along with `.` and `..` in the list of directories. If the test is run on macOS, it will fail.

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -29,7 +29,7 @@
 from .process import set_column_map
 
 # These directories are excluded from testing
-_EXCLUDE_DIRS = ['common', 's2201', 's0902', '__pycache__']
+_EXCLUDE_DIRS = ['common', 's2201', 's0902', '__pycache__', '~']
```
ajaits, line 32: is there a '~' sub-directory under census?

ajaits (approved):

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
ajaits: dissussed offline this is only needed in mac that adds ~ to dir listing.

PR #530, opened 2021-10-14
Import State and District geojsons from Datameet

pradh:
The MCF will use `isoCode` for State resolution and `indianCensusAreaCode2011` for district resolution.


thejeshgn:
This is as per census 2011. Boundaries of new districts will be missing.

pradh:
> This is as per census 2011. Boundaries of new districts will be missing. Is that okay?

I think so.  This is just something to get started with.  Once we have the new districts, we can update it.  One more thing is the resolution is through the 2011 census area code.

PR #543, opened 2021-10-22
Support for new keys in column map generator

Abilityguy:
This PR adds support for keys `measurementDenominator` and `obs_props` in the column map generator for the US Census Subject Tables.


scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
```diff
@@ -232,6 +232,13 @@ def _generate_stat_vars_from_spec(self):
         # Should that be considered an error for subject tables?
         return self.column_map
 
+    def check_obs_props(self, obs_props, stat_var):
+        """Checks for property values match in obs_props"""
+        # Check for measuredProperty
+        if obs_props['mprop'] != stat_var['measuredProperty']:
```
ajaits, line 238: what happens if 'mprop' is not present on dict obs_props or 'measuredProperty' is not present in stat_var or both are empty?

perhaps?
if obs_prop.get('mprop', 'mporp') != stat_var.get('measuredProperty', 'measuredProperty'):

scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
```diff
@@ -278,6 +278,17 @@ def _column_to_statvar(self, column):
                             )
                         stat_var[p] = v
 
+        # Handling measurementDenominator
+        if column in self.features['measurementDenominator']:
+            md_col = self.features['measurementDenominator'][column]
+            if md_col in self.column_map:
```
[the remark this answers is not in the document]
ajaits: Isn't self.column_map keys by year?

scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
Abilityguy: Thanks for pointing out this case!
I have pushed a fix.

ajaits (approved):

PR #564, opened 2021-11-17
Generate mcf file with column map

rbhoot:
The merge fixes a few bug, TODOs and adds functionality of generating mcf file along with column map.


scripts/us_census/acs5yr/subject_tables/common/data_loader.py
```diff
@@ -106,6 +103,7 @@ def process_subject_tables(table_prefix='',
                                              header_row=header_row)
 
     ## if input_path is a file, select csv/zip processing methods
+    input_path = os.path.expanduser(input_path)
```
sharadshriram, line 106: Is this a necessary change? 
`os.path.expanuser()` method is only useful if the paths are specified relative to a home directory. In this case, why do we need this extra step?
Applies to all the lines where `os.path.expanduser()` is proposed.

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
```diff
@@ -188,7 +191,8 @@ def __init__(self,
             # to be to run with --option=colmap. So rephrasing this message will help.
             print(
                 """To run this module in standalone mode, ensure the column map
-                is generated by running `generate_col_map` module.""")
+                is generated by running `generate_col_map` module or
+                process.py with --option=<colmap or all>.""")
```
sharadshriram, line 195: Can we rephrase this help message to something as follows:
"To run this module in standalone mode, ensure the column is generated first. Column map for a subject table dataset can be generated through process.py with `--option` flag set as either `colmap` or `all`. Alternatively, you can call the relevant method in `common/generate_col_map.py` module based on the input file type"

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
```diff
@@ -367,19 +371,20 @@ def _generate_mcf_from_column_map(self):
...
+            col_stat_var.append(dcid)
 
             for p, v in stat_var.items():
-                if p != 'unit' or p != 'scalingFactor':
+                if p != 'unit' and p != 'scalingFactor' and p != 'Node':
                     col_stat_var.append(f"{p}: {v}")
             # TODO: Replace the next two lines with final_mcf = final_mcf + '\n'.join(col_stat_var) + '\n\n'
-            col_stat_var = dcid + '\n'.join(col_stat_var)
-            final_mcf = final_mcf + col_stat_var + "\n\n"
+            # col_stat_var = dcid + '\n'.join(col_stat_var)
```
sharadshriram, line 386: Can we remove the `#TODO` if it has been resolved?

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
```diff
@@ -66,13 +66,10 @@ def process_subject_tables(table_prefix='',
...
-    header_row: specifies the row index to be considered as the dataframe's
-    header for column name assignment. For subject tables, the second row (row index = 1)
-    has a human readable column header. (default: 1)
-  header_row: specifies the row index to be considered as the dataframe's header
-  for column name assignment. For subject tables, the second row (row index = 1)
-  has a human readable column header. (default: 1)
+                 : since we use the 5 year subject table estimate
+  header_row: specifies the row index to be considered as the dataframe's
+            : header for column name assignment. For subject tables, the second row (row index = 1)
+            : has a human readable column header. (default: 1)
```
sharadshriram, line 72: Thanks  for cleaning this up! :)

scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
```diff
@@ -136,6 +144,53 @@ def generate_stat_var_map(spec_dict, column_list, delimiter='!!'):
...
+
+        final_mcf = final_mcf + '\n'.join(col_stat_var) + "\n\n"
+
+    print('Writing mcf output to', output_path_dir)
+    f = open(os.path.join(output_path_dir, 'statvars_all.mcf'), 'w')
+    f.write(final_mcf)
+    f.close()
+
+    return final_mcf
+
```
sharadshriram, line 192: Having the same function repeated in two places, is a bit difficult to maintain, perhaps, we can move this function to a common module?

Or better is to move this function to `process.py` since the column map is returned as a dictionary to process.py from `common/generate_col_map.py` module

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
```diff
@@ -367,19 +371,20 @@ def _generate_mcf_from_column_map(self):
...
             for p, v in stat_var.items():
-                if p != 'unit' or p != 'scalingFactor':
+                if p != 'unit' and p != 'scalingFactor' and p != 'Node':
                     col_stat_var.append(f"{p}: {v}")
             # TODO: Replace the next two lines with final_mcf = final_mcf + '\n'.join(col_stat_var) + '\n\n'
-            col_stat_var = dcid + '\n'.join(col_stat_var)
-            final_mcf = final_mcf + col_stat_var + "\n\n"
+            # col_stat_var = dcid + '\n'.join(col_stat_var)
+            final_mcf = final_mcf + '\n'.join(col_stat_var) + "\n\n"
         return final_mcf
```
sharadshriram, line 388: Having the same function repeated in two places, is a bit difficult to maintain, perhaps, we can move this function to a common module?

Or better is to move this function to `process.py` since the column map is returned as a dictionary to process.py from `common/generate_col_map.py` module

scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
```diff
@@ -84,12 +84,18 @@ def process_zip_file(zip_file_path,
         "<column-name-2>": {}, .....,
       }
   """
+    zip_file_path = os.path.expanduser(zip_file_path)
+    spec_path = os.path.expanduser(spec_path)
+    output_dir_path = os.path.expanduser(output_dir_path)
```
sharadshriram, line 89: Not sure if we need to expand paths relative to user's home directory

scripts/us_census/acs5yr/subject_tables/process.py
```diff
@@ -50,14 +54,18 @@
...
     generated_col_map = process_zip_file(input_path,
                                          spec_path,
                                          write_output=False)
+    output_dir = os.path.expanduser(output_dir)
     f = open(os.path.join(output_dir, 'column_map.json'), 'w')
     json.dump(generated_col_map, f, indent=4)
     f.close()
 
+    if create_mcf:
+        generate_mcf_from_column_map(generated_col_map, output_dir)
```
sharadshriram, line 67: Since this method is called by the option you leave `create_mcf` boolean to be True by default. Also once you bring the `generate_mcf_from_column_map` inside process.py you can call it directly after line 60.

scripts/us_census/acs5yr/subject_tables/process.py
```diff
@@ -67,8 +75,14 @@ def main(argv):
     input_path = FLAGS.input_path
     output_dir = FLAGS.output_dir
     has_percent = FLAGS.has_percent
+    create_mcf = FLAGS.create_mcf
     debug = FLAGS.debug
 
+    input_path = os.path.expanduser(input_path)
+    output_dir = os.path.expanduser(output_dir)
```
sharadshriram, line 82: Again, not sure if we need `expanduser()` here.

scripts/us_census/acs5yr/subject_tables/process.py
```diff
@@ -77,34 +91,39 @@ def main(argv):
...
-                                   input_path=input_path,
-                                   output_dir=output_dir,
-                                   column_map_path=os.path.join(
-                                       output_dir, 'column_map.json'),
-                                   spec_path=spec_path,
-                                   debug=debug,
-                                   delimiter='!!',
-                                   has_percent=has_percent)
+            if not table_prefix:
+                print('table_prefix required to run data loader')
```
sharadshriram, line 97: s/data loader/next steps of data processing.

sharadshriram (changes requested):
Thanks for the changes, Ravi!

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
rbhoot: I do use it that way.

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
rbhoot: I had kept it for reference, will remove it.

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
sharadshriram: Ok, then please have this change in `process.py` since that serves as the wrapper to these modules.

rbhoot: done!

scripts/us_census/acs5yr/subject_tables/common/data_loader.py
rbhoot: done!

scripts/us_census/acs5yr/subject_tables/common/generate_col_map.py
rbhoot: kept it only in process.py

scripts/us_census/acs5yr/subject_tables/process.py
rbhoot: didn't get this, can you explain in detail?

scripts/us_census/acs5yr/subject_tables/process.py
rbhoot: it is True by default, have added this section in case user doesn't want the mcf file to be generated.

scripts/us_census/acs5yr/subject_tables/process.py
sharadshriram: replace `data loader` with the phrase `next steps of data processing.`

PR #111, opened 2020-07-13
Import automation: progress log list resource

intrepiditee:
This PR includes code for progress log list resource associated with the endpoint /logs for creating new logs. A progress log must be linked to either a system run or an import attempt, or both. This just means that a log must be generated by either a system run or an import attempt, or both.

Thanks for reviewing.

PR #573, opened 2021-11-29
Scripts to generate nodes for elements in the periodic table

ajaits:
Uses a csv with properties for all elements and generates a csv and tmcf to create a class node for each element with some basic properties like atomic-number, symbol, type.


scripts/chemistry/periodic_table/elements.tmcf
```diff
@@ -0,0 +1,8 @@
+
+Node: E:Elements->E0
+typeOf: dcs:ChemicalElement
+name: C:Elements->Element
+chemicalName: C:Elements->Element
+chemicalSymbol: C:Elements->Symbol
+atomicNumber: C:Elements->AtomicNumber
+chemicalSubstanceType: C:Elements->Type
```
sharadshriram, line 8: @ajaits , Don't we need a dcid field in the tmcf?

ajaits: Thanks! 
Yes, adding the dcid: fixed the resolution.

scripts/chemistry/periodic_table/process.py
```diff
@@ -0,0 +1,129 @@
...
+    df = pd.read_csv(input_csv)
+
+    # Remove extra whitespaces from string columns
+    for col in ['Element', 'Symbol', 'Name', 'ChemicalName', 'Type', 'TypeOf']:
+        if col in df:
+            df[col] = df[col].str.strip()
+
+    # Add Type for periodic table without TypeOf column
+    if 'TypeOf' not in df:
+        df['TypeOf'] = 'dcs:ChemicalElement'
```
sharadshriram, line 71: I think we might want to use `dcs:Element` from #111

sharadshriram:
@ajaits, perhaps adding `ratelimit` module to the requirements.txt will fix the failing tests

pradh:
Is this appropriate to review?

scripts/chemistry/periodic_table/process.py
sharadshriram: Resolving this comment since [`ChemicalElement`](https://autopush.datacommons.org/browser/ChemicalElement) exists in the KG

PR #579, opened 2021-12-03
Boolean naming in dcid generator

Abilityguy:
Added support for Boolean naming in the dcid generator.
Picked up the convention from https://github.com/datacommonsorg/tools/blob/master/stat_var_renaming/stat_var_renaming_functions.py#L74


util/statvar_dcid_generator.py
```diff
@@ -228,6 +228,16 @@
     }
 }
 
+# This is a list of boolean properties
+_BOOLEAN_PROPS = [
```
ajaits, line 232: Are there any properties that being with words like 'is, 'has' that are either not boolean or have a different way of naming? 

If not, may be better to check if the first word (separated by camel case) is one of these and treat like a boolean property.

ajaits: looks like there are propeorties like 'hsaChildren' that are not boolean.

Would a combination of prefix and value in {true, false} siffice?

util/statvar_dcid_generator.py
```diff
@@ -458,10 +492,12 @@ def _process_constraint_property(prop: str, value: str) -> str:
     Returns:
         A string that can be used in dcid generation.
     """
-    if prop == 'naics':
+    if 'NAICS' in value:
```
beets, line 495: perhaps "startswith" here? could NAICS appear elsewhere in the name?

beets (approved):

util/statvar_dcid_generator.py
Abilityguy: There is a case where NAICS could appear elsewhere in the name.
The value could also be prepended by a namespace like `dcs: NAICS/11`.

util/statvar_dcid_generator.py
Abilityguy: There are properties which are not prefixed by an 'is' or an 'has' like `acceptsReservations` and `smokingAllowed`. However, I haven't found any naming conventions for these properties.

The `_BOOLEAN_PROPS` list contains all `is` and `has` properties from https://datacommons.org/browser/Boolean

Abilityguy: > Would a combination of prefix and value in {true, false} suffice?

Sure. However, this won't match with existing dcids like https://datacommons.org/browser/Count_Household_NoComputer

util/statvar_dcid_generator.py
beets: sounds good

beets (approved):

util/statvar_dcid_generator.py
ajaits: Ok, we can revisit generalizing this later. Pls go ahead with this set.

ajaits (approved):

PR #577, opened 2021-12-02
GeoId resolver for FBI Hate Crime Dataset

va-verma:
GeoId resolver for FBI Hate Crime Dataset
Geos : State, Country, Place

PR #580, opened 2021-12-06
Fbi hate crime

Abilityguy:
Artifacts and Scripts used for the FBI Hate Crime import.
Some of the code in this PR is borrowed from #579 and #577.

cc: @va-verma @sharadshriram @rbhoot


scripts/fbi/hate_crime/preprocess.py
```diff
@@ -0,0 +1,357 @@
...
+        incident_df,
+        groupby_cols=[],
+        agg_dict={'INCIDENT_ID': 'count'},
+        multi_index=False)
+    for idx in range(len(total_incidents)):
+        total_incidents[idx], statvars = _gen_statvar_mcf(
+            total_incidents[idx], config, population_type='CriminalIncidents')
+        statvar_list.extend(statvars)
+
+    final_df = pd.concat(total_incidents[1:])
```
rbhoot, line 239: this data is not written to csv. can we write it to incidents_count.csv?

scripts/fbi/hate_crime/preprocess.py
```diff
@@ -0,0 +1,357 @@
...
+
+# Columns which do not contribute to a constraint property value in stat var
+NONPV_COLUMNS = [
+    'MULTIPLE_OFFENSE', 'MULTIPLE_BIAS', 'INCIDENT_ID', 'DATA_YEAR',
+    'STATE_ABBR', 'AGENCY_TYPE_NAME', 'PUB_AGENCY_NAME', 'LOCATION_NAME',
+    'BIAS_AGAINST'
+]
+
+# A dict to map bias descriptions to their category
+map_dict = {
```
ajaits, line 44: Is this still needed after adding biasMotivation in the config?

Abilityguy: Yes, this is needed to generate the aggregates for different Bias Categories (race, religion, gender)
Eg: biasMotivation: gender (Sum of male and female biasMotivations).

scripts/fbi/hate_crime/preprocess.py
Abilityguy: Thanks for pointing it out.
Have made the fix. Can you take a look if it's ok?

PR #584, opened 2021-12-29
Artifacts for S2418(Median Earnings By Class Of Worker) Table

Abilityguy:
Scripts, JSON Spec and other artifacts for the ACS5YR S2418(Median Earnings By Class Of Worker) Subject Table import.
This also includes a fix for the change in decimal points when comparing two files.


scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -77,6 +78,13 @@ def _read_files(test_path, expected_path):
     return test_result, expected_result
 
 
+def _test_csv_files(test_path, expected_path, precision=3):
+    test_df = pd.read_csv(test_path).round(precision)
+    expected_df = pd.read_csv(expected_path).round(precision)
+
+    return test_df.equals(expected_df)
```
ajaits, line 85: In case equals returns false, is there  away to list out diffs as well?

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -133,6 +141,6 @@ def test_csv_mcf_column_map(self):
                     self.assertEqual(test_cmap, expected_cmap)
 
                     # Test CSV
-                    test_csv, expected_csv = _read_files(
-                        test_csv_path, paths['csv'])
-                    self.assertEqual(test_csv, expected_csv)
+                    csv_compare_bool = _test_csv_files(test_csv_path,
+                                                       paths['csv'])
+                    self.assertEqual(csv_compare_bool, True)
```
ajaits, line 146: self.assertTrue(csv_compare_bool)

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -27,7 +28,7 @@
 from .process import set_column_map
 
 # These directories are excluded from testing
-_EXCLUDE_DIRS = ['common', 's2201', '__pycache__']
+_EXCLUDE_DIRS = ['common', 's2201', 's0902', '__pycache__']
```
ajaits, line 31: pls open an issue to fix tests for 0902 if there isn't one already.

Abilityguy: Done. Have opened #585.

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
Abilityguy: Made a small change in the latest commit.
Using [assert_frame_equal](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html) now. This will throw an assertion error and points out the column where the two dataframes differ.

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
Abilityguy: Changed the assertion to `assert_frame_equal(test_df, expected_df)`.

scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -133,6 +142,6 @@ def test_csv_mcf_column_map(self):
                     self.assertEqual(test_cmap, expected_cmap)
 
                     # Test CSV
-                    test_csv, expected_csv = _read_files(
+                    test_df, expected_df = _read_csv_files(
                         test_csv_path, paths['csv'])
-                    self.assertEqual(test_csv, expected_csv)
+                    assert_frame_equal(test_df, expected_df)
```
ajaits, line 147: nice find. Can we also set check_less_precise=True so it checks for 3 digit precision.

ajaits (approved):

PR #604, opened 2022-01-24
Combine csv files to single csv file

rbhoot:


util/combine_csv.py
```diff
@@ -0,0 +1,72 @@
+from calendar import c
```
ajaits, line 1: Please add a header with:
- copyright
- how to use this utility.

rbhoot: This was for the Karnataka Demo, with some more functionalities like renaming columns in each file. I can drop it if not required.

PR #605, opened 2022-01-24
Unittest fix

rbhoot:
The test zip files have been downsized and the artifacts updated to reflect the change


scripts/us_census/acs5yr/subject_tables/subject_table_test.py
```diff
@@ -26,12 +26,11 @@
 sys.path.append(os.path.join(_SCRIPT_PATH, '.'))
 
 from common.data_loader import process_subject_tables
-from .process import set_column_map
```
rbhoot, line 29: removed this to prevent error when running ./run_tests.sh -p scripts/us_census/acs5yr/subject_tables/

Abilityguy: It seems that this removal is causing the Github tests to fail.
Could you add it back?

rbhoot: done. Thanks for pointing it out.

Abilityguy (approved):
Thanks for the fix.

PR #612, opened 2022-02-02
[Enhancement] ACS S-Table geoId resolver with more summary levels

sharadshriram:
This PR extends the summary levels for which the [resolve_geo_id.py](scripts/us_census/acs5yr/subject_tables/common/resolve_geo_id.py) module generates the place dcids from the US census GeoID strings using the FIPS code.

The summary levels that are now supported include:
|Summary level code|Summary level name|
|--------------------|---------------------|
|010|US Country-level|
|040|State-level|
|050|County-level|
|060|State-County-County Subdivision|
|140|Census tract|
|150|Block group|
|160|City/ Places|
|500|Congressional district [111th]|
|860|ZCTA|
|950,960,970|School districts|


ajaits (approved):
If there are any types of places not handled after this change, please add a note in the readme.

sharadshriram:
@ajaits, Thank you for the review. I have updated the [README](https://github.com/sharadshriram/data/blob/stable_geos/scripts/us_census/acs5yr/subject_tables/README.md) with the summary levels the module currently support.

scripts/us_census/acs5yr/subject_tables/common/resolve_geo_id.py
```diff
@@ -82,11 +83,17 @@ def convert_to_place_dcid(geoid_str):
...
-    ## Based on summary level, generate place dcid
+    ## Based on summary level and FIPS code generate place dcid
     if summary_level in _US_SUMMARY_LEVEL_GEO_PREFIX_MAP:
-        ## Update fips code
+        ## Update FIPS code
         if fips_code in _US_GEO_CODE_UPDATE_MAP:
             fips_code = _US_GEO_CODE_UPDATE_MAP[fips_code]
+
+        ## Skip resolving geoIds for "Remainder of <US State>" school districts
+        pattern = "[0-9][0-9]99999"  #check for the pattern XX9999
```
ajaits, line 93: May be faster to use fips_code.endswith('99999') rather than a re match.

Please also correct the comment to say xx99999 (5-9's)

Should this be dropped for all levels of FIPS codes or just the school district codes, 950, 960, 970.

Is there any documentation for the 99999 suffix on the US census websites?

sharadshriram: I initially started with `fips_code.endswith('99999')` but later thought the regex matching might handle this condition more flexible. I shall revert it back to use the `fips_code.endswith()` method.

This pattern has been seen only for school district summary levels codes i.e. 950, 960 and 970 and other summary levels do not seem to be affected by this so far.

I shall replace `xx99999` to `(5-9's)`.

Unfortunately, I have not come across documentation for the `99999` suffix on the US census website([for GeoIDs](https://www.census.gov/programs-surveys/geography/guidance/geo-identifiers.html))

ajaits (approved):

PR #613, opened 2022-02-02
Update packages to be python 3.9 compatible

beets:
- Added files created during script / test runs to gitignore
- Fix a few tests from the upgrade (some wrt extra precision points added during csv writing or float->string conversions)


scripts/us_eia/eia_860/test_data/2_plant.csv
```diff
@@ -1,19 +1,19 @@
...
-"eia/pp/63917","\"DC Water Solar\"","\"5000 Overlook Ave. SW, Washington, DC 20032\"","dcid:geoId/11","dcid:zip/20032","38.819244","-77.018944","NAICS/22","63917","dcid:eia/u/63572","dcid:Regulated","dcid:EIA_IndependentPowerProducer_NonCombinedHeatPower","dcid:FERC_NonQualifiedCogenerator","","dcid:FERC_QualifiedSmallPowerProducer","\"QF20-157-000\"","dcid:FERC_NonQualifiedExemptWholesaleGenerator","","","","","[69 KiloVolt]","","","dcid:False","dcid:False","dcid:False"
+"eia/pp/666","\"J D Kennedy\"","\"4215 Talleyrand Ave., Jacksonville, FL 32206\"","dcid:geoId/12","dcid:zip/32206","30.364818","-81.625967","NAICS/22","666","dcid:eia/u/9617","dcid:NonRegulated","dcid:EIA_ElectricUtility","dcid:FERC_NonQualifiedCogenerator","","dcid:FERC_NonQualifiedSmallPowerProducer","","dcid:FERC_NonQualifiedExemptWholesaleGenerator","","","","","[69.0 KiloVolt]","[138.0 KiloVolt]","[26.0 KiloVolt]","dcid:False","dcid:False","dcid:False"
```
pradh, line 18: hmm... this might be a good test for Quantity parsing, whether the resulting node with and without .0 is the same...

scripts/us_eia/eia_860/utils.py
```diff
@@ -48,7 +48,7 @@ def state_alpha2_to_dcid(alpha2: object) -> str:
 
 def zip_to_dcid(zip: object) -> str:
     if pd_types.is_number(zip):
-        return f'dcid:zip/{zip:0>5}'
+        return f'dcid:zip/{zip:0>5.0f}'
```
pradh, line 51: For future reference, could you note what changed with 3.9 that necessitates these changes?

pradh (approved):
Thanks for upgrading!

To confirm, is this expected to work with earlier versions (like 3.7)?

scripts/us_eia/eia_860/utils.py
beets: tbh i did not isolate these changes to python / pandas. i'll add a note to the PR though.

beets:
we still run our cloud tests with 3.7, so the requirements are still compatible for that.

there are some big version jumps here, so another reason to always require e2e tests...

PR #617, opened 2022-02-09
India EDM: State-wise Energy Generation and Consumption data

SenthamizhanV:
Please ignore all file changes other than those in directory `scripts/india_edm/`. Those are leakages from updating the latest base master branch.


SenthamizhanV:
@pradh Can you please share the filenames that fail the format check here? I am not getting any lint errors while running tests locally.

PR #622, opened 2022-02-11
Artifacts for Hate Crime Table 1

Abilityguy:
Artifacts and Scripts used for the FBI Hate Crime Table 1 import.


scripts/fbi/hate_crime/table1/preprocess_test.py
```diff
@@ -0,0 +1,71 @@
...
+                                 _OUTPUT_COLUMNS, preprocess._write_output_csv)
+
+            with open(cleaned_csv_path, 'r', encoding='utf-8') as f_result:
+                test_result = f_result.read()
+                expected_csv_path = os.path.join(_SCRIPT_PATH, 'testdata',
+                                                 'expected.csv')
+                with open(expected_csv_path, 'r',
+                          encoding='utf-8') as f_expected:
+                    expected_result = f_expected.read()
+                self.assertEqual(test_result, expected_result)
```
ajaits, line 71: Can do `os.remove(cleaned_csv_path)` at the end of the test to cleanup.

scripts/fbi/hate_crime/utils.py
```diff
@@ -0,0 +1,113 @@
...
+def create_mcf(stat_vars: list, mcf_file_path: str):
+    """Writes all statvars to a .mcf file."""
+    dcid_set = set()
+    final_mcf = ''
+    for sv in stat_vars:
+        statvar_mcf_list = []
+        dcid = sv['Node']
+        if dcid in dcid_set:
+            continue
+        dcid_set.add(dcid)
```
ajaits, line 95: could do `statvar_mcf_list = [f'Node: dcid:{dcid}')]` here to avoid an extra concat in line 99

scripts/fbi/hate_crime/utils.py
```diff
@@ -0,0 +1,113 @@
...
+        for p, v in sv.items():
+            if p != 'Node':
+                statvar_mcf_list.append(f'{p}: dcs:{v}')
+        statvar_mcf = 'Node: dcid:' + dcid + '\n' + '\n'.join(statvar_mcf_list)
+        final_mcf += statvar_mcf + '\n\n'
```
ajaits, line 100: Can write each node to file as it is constructed without storing all into a local string.

scripts/fbi/hate_crime/README.md
```diff
@@ -0,0 +1,43 @@
...
+To generate `cleaned.csv`, `output.mcf` for a publication. run:
+
+```bash
+cd table<publication number>
+python3 preprocess.py
+```
+
+### Data Caveats:
+- New Jersey is missing data for year 2012 in publications 11 and 12
+- Data for a few locations of crime (publication 10) are missing for certain years
```
ajaits, line 15: Please add a note if the output.mcf is not to be used and instead the statvars definition comes from the aggregations scripts due to the change in populationType, etc.

ajaits (commented):
Thank you for the well documented code!

scripts/fbi/hate_crime/table1/preprocess_test.py
Abilityguy: Since `cleaned_csv_path` is in the temporary directory, the file should be cleaned once it comes out of the context manager block. [Docs](https://docs.python.org/3.9/library/tempfile.html#tempfile.TemporaryDirectory)

scripts/fbi/hate_crime/README.md
Abilityguy: Done.

scripts/fbi/hate_crime/utils.py
Abilityguy: Done.

scripts/fbi/hate_crime/utils.py
Abilityguy: Done.

scripts/fbi/hate_crime/table1/preprocess_test.py
ajaits: Ah! thanks. In that case no cleanup needed.

ajaits (approved):

PR #626, opened 2022-02-16
India/LGD - Sync LGD Districts with Wikidata Districts

thejeshgn:
- Initially, wikidata had `New Delhi` [Q987](https://www.wikidata.org/wiki/Q987) as a district. Hence it was mapped. Now they have created a new district entity (on Dec 2021.) `New Delhi district`  [Q110067863](https://www.wikidata.org/wiki/Q110067863). Hence the mapping change. Check [revision - Q110067863](https://www.wikidata.org/w/index.php?title=Q110067863&oldid=1542627512) and [revision - Q987](https://www.wikidata.org/w/index.php?title=Q987&oldid=1542627675).
> - [ ] @SenthamizhanV see if 👆  affects any of your import.   @pradh 
- [Q620801](http://www.wikidata.org/entity/Q620801) - Wikidata has updated the name to Narmadapuram . But LGD is yet to update the name, even though the official name is now https://hoshangabad.nic.in/en/about-district/.
- [Q101088203](http://www.wikidata.org/entity/Q101088203) Added Balaji - District from Assam
- [Q307474](http://www.wikidata.org/entity/Q307474) Added Purulia - District of WB

To generate the final CSV  `LocalGovernmentDirectory_Districts.csv` run 

```bash
python -m india_lgd.local_government_directory_districts.preprocess
```

Please note the raw CSVs are not automatically downloaded. They are manually downloaded as mentioned in the readme file.  LGD has captcha. 

CC:  @sharadshriram


scripts/india_lgd/local_government_directory_districts/preprocess.py
```diff
@@ -119,9 +119,9 @@ def _load_and_format_lgd(self):
         # if there are NA values then replace it with '' character
         self.lgd_df = pd.read_csv(self.lgd_csv, dtype=str)
         self.lgd_df.fillna('', inplace=True)
-        # Drop title rows in the top and empty rows after 740.
-        # The actual data is between 2nd and 740th row. So keep only them.
-        self.lgd_df = self.lgd_df.iloc[1:739]
+        # Drop title rows in the top and empty rows after 741.
+        # The actual data is between 2nd and 741st row. So keep only them.
+        self.lgd_df = self.lgd_df.iloc[1:741]
```
beets, line 124: is there a way to future proof this (so we won't have to keep updating these constants, which someone else might forget to do)..

- perhaps you could call read_csv with header=None or header=2
- i'm also surprised there are empty rows - the csv looks fairly clean and doesn't include empty lines (or you could use skip_blank_lines in read_csv).

beets (commented):
thanks for the change.

if the csv updates were generated programatically, could you add what you ran to generate this to the PR description?

scripts/india_lgd/local_government_directory_districts/preprocess.py
thejeshgn: @beets  Thank you. Makes sense. Done.

beets (approved):
thanks for the updates!

thejeshgn:
@ajaits  can we merge this?

PR #669, opened 2022-04-20
ACS5YrSubjectTable: S2703 Artifacts

Abilityguy:
Includes the S2703_spec.json and tests for the S2703 (Private Health Insurance Coverage By Type) Subject Table import.


scripts/us_census/acs5yr/subject_tables/S2703/S2703_spec.json
```diff
@@ -0,0 +1,219 @@
+{
+    "preprocess": {
+        "find_and_replace": {
+            "Estimate!!Total!!Civilian noninstitutionalized population!!COVERAGE ALONE OR IN COMBINATION!!Employer-based health insurance alone or in combination": "Estimate!!Total!!Civilian noninstitutionalized population",
```
ajaits, line 4: Is the token `Employer-based health insurance alone or in combination` being dropped?

This has a PV mapping in the confing below:
```
         "Employer-based health insurance alone or in combination": "WithEmployerBasedHealthInsurance"
```

Abilityguy: The token is only removed for the `Total` columns where the health insurance PV does not apply.  The token `Employer-based health insurance alone or in combination` applies for the `Private Coverage` set of columns.

Screen Shot 2022-05-21 at 10 11 33 AM

ajaits (approved):

PR #671, opened 2022-04-27
SV Groups corresponding to Categories for WHO GHO indicators. Source:…

jehangiramjad:
… https://www.who.int/data/gho/data/indicators


pradh (approved):
Thank you!  It might be useful to have a README in scripts/who/gho with a pointer to the source hierarchy, and a few words on how you compiled the .csv file?

Just enough info for someone in future to know how we got here, in case they need to retrace the steps...

PR #685, opened 2022-05-13
Remove .gitattributes at root for git LFS

ajaits:
Restored small csv test files from git LFS to normal csv:

```
for file in $(git lfs ls-files | grep test | cut -d\* -f2); do
   cp $file tmp/
   git lfs untrack $file
   git rm -cached $file
   git commit -m "remove git LFS files"
   git push -u origin gitlfs
   git add -f $file
   cp tmp/$(basename $file) $file
done
git commit -m "restore original files"
git push -u origin gitlfs
```


beets (approved):
thanks for the fix!

jehangiramjad (approved):

PR #713, opened 2022-06-28
[feature] mapping for US State/Division names to dcid

sharadshriram:
This PR adds a mapping to resolve the US State names and US Census Division names to their respective `dcid`.

cc. @ajaits

PR #714, opened 2022-06-28
[CDC-NORS] Non-infectious disease import artifacts

sharadshriram:
This PR adds the processing script and test import artifacts generated for the NORS Non-infectious disease import.

The test to validate the statvar mcf file is flaky -- since the order of nodes in the expected file and generated file differ

This PR depends on #713 for scripts to work
This PR supersedes #708 -- which did not sync to the changes

cc: @ajaits


ajaits (commented):
Can you pls move this to git LFS:
scripts/us_cdc/non_infectious_diseases/data/NationalOutbreakPublicDataTool.xlsx

ajaits (approved):

PR #728, opened 2022-07-13
Add some docs on formatting DC strings and repeated values

pradh:


spiekos (approved):
This is useful information and is easy to follow. Thanks for adding this Prashanth!

PR #757, opened 2022-10-05
Some place name resolver fixes

pradh:
1. Do batching when calling our recon API
2. Handle cases where multiple rows return the same placeId


beets (approved):

pradh:
Thanks!

PR #758, opened 2022-10-06
Add GPCC SPI import scripts

Fructokinase:


pradh (commented):
Please include a README

scripts/noaa/gpcc_spi/download.py
```diff
@@ -0,0 +1,79 @@
...
+                    "Part of the download failed. Please re-run or fix the script."
+                )
+
+
+def main(_):
+    os.makedirs(FLAGS.output_dir, exist_ok=True)
+
+    periods = FLAGS.periods
+    if not periods:
+        periods = ['1', '2', '3', '6', '9', '12', '24', '36', '48', '60', '72']
```
pradh, line 74: Should we limit this to what we import?

scripts/noaa/gpcc_spi/create_test_nc.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2022 Google LLC
```
pradh, line 1: Move this to under test/.  Its a onetime helper script?

scripts/noaa/gpcc_spi/download.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2022 Google LLC
```
pradh, line 1: We should have a wrapper shell script or at least README instructions that ties everything together.

scripts/noaa/gpcc_spi/gpcc_spi_aggregation.py
```diff
@@ -0,0 +1,130 @@
...
+import glob
+from pathlib import Path
+
+from absl import flags
+from absl import app
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string('in_pattern', '/tmp/gpcc_spi/out/*.csv',
+                    'The directory the csvs are.')
```
pradh, line 31: What are the CSVs?  Wonder why not have the same program generate both grids and aggregations?

So its easier to run them as a single cron.

scripts/noaa/gpcc_spi/gpcc_spi_aggregation.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def _read_spi_data_csv(csv_path):
+
+    def grid_to_coord(grid):
+        """grid_1/lat_lng -> lat^lng"""
+        lat_lng = grid.split('/')[1]
+        return "^".join(lat_lng.split('_'))
+
+    df = pd.read_csv(csv_path, usecols=['time', 'place', 'spi'])
```
pradh, line 92: How big are the CSVs?  For large ones I tend to prefer not doing pandas and instead looping over rows.

pradh (commented):
Code looks clean!  We should add documentation and see if we can simplify the flow with a single script

scripts/noaa/gpcc_spi/gpcc_spi_aggregation.py
Fructokinase: Was hoping to isolate csv preprocessing and aggregation each as isolated tasks. Isolated tasks are easier to orchestrate and task level retry is easier to implement. 

Other datasets may have several layers of preprocessing and having everything in the same script might not generalize well when we need to refactor.

To your point of running it as a single cron, I added a wrapper script which does that for now. It's just slower to do them in series but as long as we only need to process small amounts of datas in refresh, speed perhaps wouldn't be a limiting factor.

PR #762, opened 2022-10-13
RFF: additional Climate Variability SVs with historical data at US county-level

pradh:


scripts/rff/historical_weather_variability/county/R/county_aggregation.R
```diff
@@ -0,0 +1,248 @@
+### Aggregate PRISM to county ###
+rm(list=setdiff(ls(), c("master_path", "code_path")))
+path <- master_path
+
+coords <- read.csv(paste0(path, "replication/data/monthly_dataframes/month_ppt_coordinates.csv"))
```
jehangiramjad, line 5: There are several string and integer constants being used extensively. Suggestion is to use variables with the constants so that it is easier to understand and follow?

scripts/rff/historical_weather_variability/county/R/county_aggregation.R
```diff
@@ -0,0 +1,248 @@
...
+
+coords$fips <- over$GEOID
+
+write.csv(coords, paste0(path, "replication/data/prism_coordinates_fips.csv"), row.names = F)
+
+
+#### RANGES ####
+
+# TMIN
+load(paste0(path, "replication/data/ranges/tmin/tmin_ranges_dataframe.Rdata"))
```
jehangiramjad, line 24: For this and the sections below, it might be better to abstract the procedure in separate helper functions. Eventually, after every call to a helper function returns, the merge() function can be used..

scripts/rff/historical_weather_variability/county/R/heatwave_5year_rasters.R
```diff
@@ -0,0 +1,119 @@
...
+
+
+path <- master_path 
+setwd(path)
+
+# count 
+load(paste0(path, "replication/data/heatwaves/hw_count_dataframe.Rdata"))
+ref <- raster(paste0(path, "replication/data/heatwaves/4km/annual/hw_count_1981.tif"))
+
+data$count_1981_1985 = data$hw_count_1981 + data$hw_count_1982 +
```
jehangiramjad, line 11: Might be easier to abstract these away in a function and access the column names programmatically. In R, I believe one can assign column name strings to variables and then access them: https://stackoverflow.com/questions/19730806/access-data-frame-column-using-variable

scripts/rff/historical_weather_variability/county/R/heatwave_5year_rasters.R
```diff
@@ -0,0 +1,119 @@
...
+data$count_2006_2010 = data$hw_count_2006 + data$hw_count_2007 + 
+  data$hw_count_2008 + data$hw_count_2009 + data$hw_count_2010
+
+data$count_2011_2015 = data$hw_count_2011 + data$hw_count_2012 + 
+  data$hw_count_2013 + data$hw_count_2014 + data$hw_count_2015
+
+data$count_2016_2020 = data$hw_count_2016 + data$hw_count_2017 + 
+  data$hw_count_2018 + data$hw_count_2019 + data$hw_count_2020
+
+data <- data[,c(1,2,43:50)]
```
jehangiramjad, line 35: Another instance of using integers which could benefit from helpful comments or global variables etc.

scripts/rff/historical_weather_variability/county/R/heatwave_rasters_to_df.R
```diff
@@ -0,0 +1,65 @@
...
+
+colnames(data) <-  c("x", "y", paste0("hw_count_", 1981:2020))
+save(data, file =paste0(dir, "replication/data/heatwaves/hw_count_dataframe.Rdata"))
+
+
+# HW LENGTH
+path <- paste0(dir, "replication/data/heatwaves/4km/annual")
+
+files <- list.files(path)
+files <- files[substr(files, 1,9)=="hw_length"]
```
jehangiramjad, line 28: Seems a bit fragile/brittle. Helpful comments, e.g. why should nchar(files) be 18?

jehangiramjad (approved):
Approving this just in case it is a blocker. However, a couple of high level comments which are applicable in several places:

1. Using string/integer constants should either be done via global/file constants or have useful comments to help with readability and debugging
2. Helper functions could be super useful  instead of the sections in the scripts/files.
3. In R, column names can be assigned to variables. This means several hardcoded column names can be programmatically accessed to avoid unintended bugs etc.

pradh:
Thanks for the comments Jehangir!  I've conveyed this to the author of the code who thinks the comments all make sense and plans to address them after next week.  For tracking, I've filed an issue - https://github.com/datacommonsorg/data/issues/765

scripts/rff/historical_weather_variability/county/R/county_aggregation.R
pradh: Marking resolved to submit.

scripts/rff/historical_weather_variability/county/R/county_aggregation.R
pradh: Marking resolved to submit.

scripts/rff/historical_weather_variability/county/R/heatwave_5year_rasters.R
pradh: Marking resolved to submit.

scripts/rff/historical_weather_variability/county/R/heatwave_5year_rasters.R
pradh: Marking resolved to submit.

PR #774, opened 2022-11-18
Utility classes for config and counters

ajaits:
Common utility libraries to load config parameters from dict and counters.

Copied from PR#764


ajaits:
Created PR 775

PR #779, opened 2022-11-29
Batch recon calls

Fructokinase:
This is a change split from the earthquake preprocessing PR  https://github.com/datacommonsorg/data/pull/751

For large number of coordinates that need to be resolved, the service will fail without batching. Batch size of 50 was eyeballed. Also add a retry in case of 5xx errors, which we see sometimes.


util/latlng_recon_service.py
```diff
@@ -16,13 +16,48 @@
 See latlng_recon_service_test.py for usage example.
 """
 
+import concurrent.futures
+import logging
 import requests
+from typing import Callable, Dict, List, NewType, TypeVar, Tuple
 
-_RECON_ROOT = 'https://api.datacommons.org/v1/recon/resolve/coordinate'
+LatLng = NewType('LatLng', Tuple[float, float])
```
beets, line 24: Since there isn't much guidance on this, I wonder if we should use the suffix Type for these to distinguish from vars / classes (e.g. LatLngType, etc). Wdyt?

Nit: DCID --> DcidType.

beets: We should also make these module private for now (add an underscore prefix).

beets (commented):
thanks for splitting these out

util/latlng_recon_service.py
Fructokinase: >  I wonder if we should use the suffix Type 

Good idea, added Type suffix. Since some of these types are used for function arguments, which are called externally, they need to be exposed as opposed to being private?

beets: Ah got it. Ok, then it's better if exposed with the Type suffix.

beets (approved):

PR #796, opened 2023-01-03
Filter importing script to only import data post october to make auto refresh manageable.

pulkit-s:
To get around issue with refreshing fire data b/263500910.

We have data manifested till 2022-10-10. This change will only add data post that data to the refresh csv.


shifucun (commented):
Wonder how would this impact the generated mcf files?

pulkit-s:
> 

my plan is to move data till october 2022 to a historical folder in colossus and have a current folder for the refreshed data. 

Then in the mcf url [in the manifest](https://source.corp.google.com/piper///depot/google3/datacommons/import/mcf/manifest/prod_us_stats_mcfs.textproto;rcl=499315849;l=491), we can use something like /cns/jv-d/home/datcom/v3_resolved_mcf/fire/wfigs/*/nodes.mcf@1 where the asterisk expands to historical and current. Do you think that would work?

shifucun:
> > 
> 
> my plan is to move data till october 2022 to a historical folder in colossus and have a current folder for the refreshed data.
> 
> Then in the mcf url [in the manifest](https://source.corp.google.com/piper///depot/google3/datacommons/import/mcf/manifest/prod_us_stats_mcfs.textproto;rcl=499315849;l=491), we can use something like /cns/jv-d/home/datcom/v3_resolved_mcf/fire/wfigs/*/nodes.mcf@1 where the asterisk expands to historical and current. Do you think that would work?

If we have such usage now, then it sounds good to me. @Spaceenter any suggestions?

scripts/fires/wfigs_data.py
```diff
@@ -47,7 +47,9 @@
...
+# FIRE_LOCATIONS_URL = "https://services3.arcgis.com/T4QMspbfLg3qTGWY/arcgis/rest/services/Fire_History_Locations_Public/FeatureServer/0/query?where=1%3D1&returnGeometry=false&outFields=InitialLatitude,InitialLongitude,InitialResponseAcres,InitialResponseDateTime,UniqueFireIdentifier,IncidentName,IncidentTypeCategory,IrwinID,FireCauseSpecific,FireCauseGeneral,FireCause,FireDiscoveryDateTime,ContainmentDateTime,ControlDateTime,IsCpxChild,CpxID,DailyAcres,POOFips,POOState,EstimatedCostToDate,TotalIncidentPersonnel,UniqueFireIdentifier&outSR=4326&orderByFields=FireDiscoveryDateTime&f=json&resultType=standard"
+# Only download data from Oct 10, 2022 to make auto refresh manageable.
+POST_OCT_2022_FIRE_LOCATIONS_URL = "https://services3.arcgis.com/T4QMspbfLg3qTGWY/arcgis/rest/services/Fire_History_Locations_Public/FeatureServer/0/query?where=FireDiscoveryDateTime>'2022-10-10'&outFields=*&outSR=4326&f=json"
```
shifucun, line 52: why is this "2022-10-10" instead of "2022-10-01"?

pulkit-s: We have manifested data till 2022-10-10 for now, so I am making the size here as manageable as possible by only downloading data after that.

Spaceenter (approved):

scripts/fires/wfigs_data.py
shifucun: Got it. Thanks. Worth to note in the PR description.

pulkit-s: Done

PR #816, opened 2023-02-13
[Census TMCF] Replacing '&' with '++' to allow DC frontend to parse the DCID correctly

jehangiramjad:
Leaving the '&' in the opaque mappings meant that while the cache got all the data, the DC frontend was unable to show the data due to the '&' making the dcid be split assuming it was a url.

After a discussion with @pradh, decided to replace '&'. 

Also: apologies for not having deleted the branch successfully before this PR (it still shows the commit history from the prior PR--the only changes here are in the test files and the one liner for replacing '&' with '++').


scripts/us_census/enhanced_tmcf/process_etmcf.py
```diff
@@ -128,6 +128,7 @@ def _process_var_measured_new_column(
 
     # TODO: remove these checks when a better format for opaque mapping is agreed.
     query_format = query_format.replace("?", "__")
+    query_format = query_format.replace("&", "//")
```
beets, line 131: maybe `~~` (something not in the URL spec)

jehangiramjad: ++ works so trying that

pradh (approved):

PR #819, opened 2023-02-14
[Census TMCF] Tweaking the expected template to enable SV Names appear

jehangiramjad:


scripts/us_census/enhanced_tmcf/sample_input_etmcf.tmcf
```diff
@@ -11,44 +11,79 @@ footnote: C:T->GEO_ID_F
...
-name: Number of establishments
 value: C:T->ESTAB
-variableMeasured: T:ESTAB?naics2012=C:T->NAICS2012&taxstat=C:T->TAXSTAT&typop=C:T->TYPOP
+variableMeasured: E:T->E1.5
 observationAbout: E:T->E0
 observationDate: C:T->YEAR
 
+Node: E:T->E1.5
+typeOf: dcs:StatisticalVariable
+statType: measurementResult
```
pradh, line 21: Sorry, this should not be measurementResult for numeric values.  We should change this to measuredValue.  We sort of drop measurementResult stuff from the caches / SV hierarchy (https://source.corp.google.com/search?q=kMcfMeasurementResult), so currently it doesn't show up.

PR #834, opened 2023-03-23
ICTV Import

spiekos:
Note: need to add Notes and Caveats as well as commands for running and testing scripts. Also need to add tests and testing files


scripts/biomedical/ICTV_Taxonomy/scripts/tests.sh
```diff
@@ -0,0 +1,31 @@
...
+#!/bin/bash
+
+java -jar /Applications/datacommons-import-tool-0.1-alpha.1-jar-with-dependencies.jar lint tMCFs/virusMasterSpeciesList.tmcf CSVs/VirusSpecies.csv ICTV*.mcf
+mv dc_generated species
+
+java -jar /Applications/datacommons-import-tool-0.1-alpha.1-jar-with-dependencies.jar lint tMCFs/virusTaxonomy.tmcf CSVs/VirusIsolates.csv ICTV*.mcf
+mv dc_generated virus_isolates
+
+java -jar /Applications/datacommons-import-tool-0.1-alpha.1-jar-with-dependencies.jar lint tMCFs/virusGenomeSegment.tmcf CSVs/VirusGenomeSegments.csv ICTV*.mcf
+mv dc_generated genome_segments
```
pradh, line 31: I think this script is very specific to where you downloaded the file to.

Maybe as the first step, you could download `https://github.com/datacommonsorg/import/releases/download/0.1-alpha.1k/datacommons-import-tool-0.1-alpha.1-jar-with-dependencies.jar` to `/tmp` and then run it?

It also assumes the user has `java` JRE installed, which you can point out in the comments maybe.

scripts/biomedical/ICTV_Taxonomy/scripts/create_virus_taxonomic_ranking_enums.py
```diff
@@ -0,0 +1,175 @@
...
+"""
+
+# load environment
+import pandas as pd
+import sys
+
+
+# declare universal variables
+HEADER = [
+'sort',\
```
pradh, line 31: nit: I don't think you need `\` for lists etc that are delimited by `,`

scripts/biomedical/ICTV_Taxonomy/README.md
```diff
@@ -0,0 +1,161 @@
+
+# Importing Master Species List and Virus Metadata Resource from the International Committee on Taxonomy of Viruses (ICTV)
```
pradh, line 2: Could you please check the test failures in https://github.com/datacommonsorg/data/pull/834/checks?check_run_id=22280933775

I think you'll need to fix lint (`./run_tests.sh -f`)

pradh (commented):
Thanks for the detailed doc!

I only skimmed through the tests, one main comment on the validation script.

scripts/biomedical/ICTV_Taxonomy/scripts/create_virus_taxonomic_ranking_enums.py
spiekos: removed from all three python script files

scripts/biomedical/ICTV_Taxonomy/scripts/tests.sh
spiekos: updated the tests script so downloading the tool is part of the script and the comments in the README.md to explain this functionality in the tests subsection

pradh (approved):
Thanks for the updates!

scripts/biomedical/ICTV_Taxonomy/README.md
spiekos: ran the lint tests!

PR #862, opened 2023-06-29
Keeping the DCIDs of Warangal and Malerkotla same, even if wikidataId has changed.

thejeshgn:
Keeping the DCIDs of Warangal and Malerkotla same, even if wikidataId has changed. But updating the wikidataId attribute.

Refer - https://github.com/datacommonsorg/data/pull/743

This exception is required only for Warangal and Malerkotla.

|District           |WikiDataID Previous|Exists on DC|WikiDataID New|Exists on DC|Map to the existing DC Node                                     |
|-------------------|-----------------|------------|----------------|------------|----------------------------------------------------------------|
|Warangal District  |Q28169759        |Yes         |Q15399          |No          |Keep the same DCID and update the node with wikidataId          |
|New Delhi District |Q110067863       |No          |Q8560886        |Yes         |Update the Node with just LGD code. No special care is required.|
|Malerkotla District|Q1470987         |Yes         |Q107016021      |No          |Keep the same DCID and update the node with wikidataId          |

OLD PR: https://github.com/datacommonsorg/data/pull/759

PR #863, opened 2023-07-07
Updated the india geo lookup utility functions

thejeshgn:
- Updated the look up CSV
- Added more tests

Reference: https://github.com/datacommonsorg/data/pull/862


scripts/india/geo/districts_test.py
```diff
@@ -58,3 +58,43 @@ def test_get_district_name_to_lgd_code_mapping(self):
         district_lgd_code = mapper.get_district_name_to_lgd_code_mapping(
             "telangana", "warangal")
         self.assertEqual(district_lgd_code, "522")
+
+        district_lgd_code = mapper.get_district_name_to_lgd_code_mapping(
+            "andhra pradesh", "nandyal")
```
ajaits, line 63: Does the state lookup support queries like 'Andhrapradesh' or 'AP' ?
Should INDIA_STATES_LGD_CODES be expanded with more versions of the state names?

ajaits (approved):

scripts/india/geo/districts_test.py
```diff
@@ -58,3 +58,43 @@ def test_get_district_name_to_lgd_code_mapping(self):
...
+        self.assertEqual(district_lgd_code, "755")
+
+        district_lgd_code = mapper.get_district_name_to_lgd_code_mapping(
+            "andhra pradesh", "ntr")
+        self.assertEqual(district_lgd_code, "749")
+
+        district_lgd_code = mapper.get_district_name_to_lgd_code_mapping(
+            "Meghalaya", "eastern west khasi hills")
```
SenthamizhanV, line 71: nit: "meghalaya" with lowercase 'm'

SenthamizhanV (approved):

scripts/india/geo/districts_test.py
thejeshgn: Its purposely done in a testcase to see if it matches regardless of Case.

scripts/india/geo/districts_test.py
thejeshgn: @ajaits  We can add more look up formats. I will create an issue and we can add them as and when required.

PR #867, opened 2023-07-14
OECD import scripts

n-h-diaz:
this is just to check something in, but I'll iteratively improve the scripts + add tests


pradh (approved):
Thanks!  please consider expanding on the README with something on SDMX :)

PR #870, opened 2023-07-27
add latest agriculture stat var mappings

n-h-diaz:
(note: this is just for reference in future refreshes and does not work with the existing script)


scripts/us_usda/nass_census/sv.csv
```diff
@@ -0,0 +1,159 @@
+name,sv,unit
```
keyurva, line 1: Based on the data I'm getting from the api, many of them don't have statvar mappings here (e.g. "FOOD FISH, CATFISH, WATER, IN PRODUCTION - ACRES OF WATER"). How do you propose we deal with those?

keyurva (commented):
Since we'll be using the new script for both survey and census, does it make sense to put these files in a new folder (say `usda/quickstats` or something similar)?

scripts/us_usda/nass_census/sv.csv
keyurva: You can see all short descriptions they support here: https://quickstats.nass.usda.gov/api/get_param_values?key=569256A0-9CF4-34F0-86F3-FC477003330A&param=short_desc

n-h-diaz:
> Since we'll be using the new script for both survey and census, does it make sense to put these files in a new folder (say `usda/quickstats` or something similar)?

good idea, moved to new folder

scripts/us_usda/nass_census/sv.csv
n-h-diaz: I think ideally we add mappings for these (and add corresponding schema for them), but if we just want to refresh the existing variables, we can skip the missing ones (this list only contains a veryyyyy small subset of agriculture data )

keyurva: I'll leave that up to you and @pradh since I don't have enough context yet. For now, perhaps add a TODO to add missing mappings.

keyurva (approved):

scripts/us_usda/nass_census/sv.csv
n-h-diaz: added todo

PR #889, opened 2023-08-11
Add support for extracting and writing all WB indicator codes to a CSV.

keyurva:


scripts/world_bank/datasets/datasets.py
```diff
@@ -38,10 +47,22 @@
...
 from urllib3.util.ssl_ import create_urllib3_context
 from absl import flags
+import zipfile
+import codecs
 
 FLAGS = flags.FLAGS
 
-flags.DEFINE_boolean('download_datasets', False, 'Downloads the datasets.')
+
+class Mode:
```
ajaits, line 56: We can use Python Enums:
https://docs.python.org/3/library/enum.html

Then the command line flags can be `--mode=FETCH_DATASETS` without the `mode.` prefix

ajaits (approved):

scripts/world_bank/datasets/datasets.py
keyurva: Using enum classes with flags is a bit unwieldy (see b/72756738) so went with strings. Node that users don't have to specify the prefix. They only need to specify `--mode=fetch_datasets`

PR #957, opened 2023-12-12
US BLS Jolts data refresh

swethammkumari:
PR request is created for US BLS Jolts .
The existing dcid's are retained by modifying the script. 
If any new SV introduces in future , that should be added to the dictionaries that are in map_config.py & mcf_config.py .


hareesh-ms (commented):
Instead of creating a separate sub directory for updated files, can you modify the original set of files.  Also include unittest for this.

swethammkumari:
> Instead of creating a separate sub directory for updated files, can you modify the original set of files. Also include unittest for this.

Creating a new PR as the old one was pushed incorrectly.

PR #969, opened 2024-01-23
Import Automation: Zero day vulnerability force upgrade libwebp version

jehangiramjad:
We continue to be notified that the Docker images are still using libwebp 1.2.4-0.2. Trying with a force version upgrade.

### DIFF
--- a/import-automation/executor/Dockerfile
+++ b/import-automation/executor/Dockerfile
@@ -17,6 +17,10 @@
 FROM python:3.11.4
 
 RUN apt-get update \
+    # Ensure a fix against a zero-day vulnerability
+    # CVE-2023-4863.
+    # Ref: https://snyk.io/blog/find-and-fix-webp-vulnerability-cve-2023-4863/#upgrade
+    && apt-get satisfy "libwebp-dev (>= 1.3.2)"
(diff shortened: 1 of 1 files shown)


shifucun (commented):
In this link, you can see each docker image and it's vul details. Looks like the most recent one does not reduce it.

https://pantheon.corp.google.com/gcr/images/datcom-ci/global/datacommons-import-automation-executor?mods=-monitoring_api_staging

Also if you click into the most recent one, looks like there are no fixes for the found issues. So maybe double check the reported vul and the one listed here.

shifucun:
> In this link, you can see each docker image and it's vul details. Looks like the most recent one does not reduce it.
> 
> https://pantheon.corp.google.com/gcr/images/datcom-ci/global/datacommons-import-automation-executor?mods=-monitoring_api_staging
> 
> Also if you click into the most recent one, looks like there are no fixes for the found issues. So maybe double check the reported vul and the one listed here.

Note the reopened one in b/320239641 refers to an old docker image (which i think is still in deployment in gke). So the latest one should be ok.

jehangiramjad:
Looks like this isn't needed and that the vulnerability reported is now gone.

PR #972, opened 2024-01-31
Remove frozendict from us_fed script

ajaits:
Remove frozendict import as it generates an error in python 3.11:
```python generate_csv_and_mcf.py 
Traceback (most recent call last):
  File "/usr/local/google/dc/refresh/git_us_usda/data/scripts/us_fed/treasury_constant_maturity_rates/generate_csv_and_mcf.py", line 31, in <module>
    from frozendict import frozendict
  File "/google/src/cloud/ajaits/dc-github/google3/env_py_3_11/lib/python3.11/site-packages/frozendict/__init__.py", line 16, in <module>
    class frozendict(collections.Mapping):
                     ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'collections' has no attribute 'Mapping'
```


jehangiramjad (approved):
Thanks!

PR #977, opened 2024-02-03
[Import Automation] Script to invoke updates for imports locally

jehangiramjad:
In this PR, we have:

1. A new script to trigger updates for imports.
2. Print some diagnostics at completion time
3. Use currently checked out repo instead of using the latest committed master branch (which happens via the Cloud Scheduler workflow).

Coming next:

- Ability to parse and set/override configs from a local file (including user script args)
- Similar flow for Scheduling imports on the Cloud Scheduler.
- More diagnostics.

### DIFF
--- a/import-automation/executor/README.md
+++ b/import-automation/executor/README.md
@@ -47,12 +47,34 @@ Commons knowledge graph using the importer.
 
 ## Running locally
 
+## Updating An Import Locally
+
+Authenticate with GCP first: `gcloud auth application-default login`
+
+You can execute an import job from your local end by invoking the following script:
+
+```
+Run `./schedule_update_import.sh -u <config_project_id> <path_to_import> <import_script_args>`
+```
+
+Run `./schedule_update_import.sh --help` for usage.
+
+`<config_project_id>` is the GCP project id where the config file is stored, e.g. `datcom-import-automation`.
+`<path_to_import>` is the path to the import (relative to the root directory of the `data` repo), with the name of the import provided with a colon, e.g. `scripts/us_usda/quickstats:UsdaAgSurvey`.
+`import_script_args` is one string (enclosed by quotes) representing all the command line args to be provided to the import script, e.g. `"--start_year=2023 --another_flag=val"`.
+
+Example invokation:
+
+```
+Run `./schedule_update_import.sh -u datcom-import-automation scripts/us_usda/quickstats:UsdaAgSurvey "--start_year=2023"`
+```
+
+## Local Executor [should be deprecated soon]
+
 ```
 PYTHONPATH=$(pwd) python app/main.py
 
 ``
-
-## Local Executor
 
 Run `. run_local_executor.sh --help` for usage.
 
--- a/import-automation/executor/README.md
+++ b/import-automation/executor/README.md
@@ -58,19 +58,18 @@ Once the script runs to completion, the data directory's latest update is printe
 Ensure this script is executed from the directory which contains `schedule_update_import.sh`, i.e. from `/data/import-automation/executor`.
 
 ```
-Run `./schedule_update_import.sh -u <config_project_id> <path_to_import> <import_script_args>`
+Run `./schedule_update_import.sh -u <config_project_id> <path_to_import>`
 ```
 
 Run `./schedule_update_import.sh --help` for usage.
 
 `<config_project_id>` is the GCP project id where the config file is stored, e.g. `datcom-import-automation`.
 `<path_to_import>` is the path to the import (relative to the root directory of the `data` repo), with the name of the import provided with a colon, e.g. `scripts/us_usda/quickstats:UsdaAgSurvey`.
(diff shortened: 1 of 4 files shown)


import-automation/executor/schedule_update_import.sh
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+function help {
+  echo "#Usage: -us <config_project_id> <absolute_import_path> <import_script_args>"
+  echo "## <config_project_id> is the GCP project ID where the config file is located." 
+  echo "## Update an import specified by <absolute_import_path>, e.g. scripts/us_usda/quickstats:UsdaAgSurvey"
+  echo "## Provide all args for the import script as one string, i.e. with quotes, e.g. \"--flag1=val1 --flag2=val2\""
+}
```
ajaits, line 21: `exit 1` to terminate if args are not correct?

import-automation/executor/schedule_update_import.py
```diff
@@ -0,0 +1,234 @@
...
+    # Converting string to list
+    args_list = import_script_args.split(' ')
+    if type(args_list) != type([]):
+        raise Exception(
+            'Flag: import_script_args could not be parsed into a list.')
+
+    # Get the root repo directory (data). Assumption is that this script is being
+    # called from a path within the data repo.
+    cwd = os.getcwd()
+    repo_dir = cwd.split("data")[0] + "data"
```
ajaits, line 193: Would it be useful to have a way to run this script to execute from repo rather than local dir always?

import-automation/executor/schedule_update_import.py
```diff
@@ -0,0 +1,234 @@
...
+    logging.info(f'{mode} called with the following:')
+    logging.info(f'Config Project ID: {_FLAGS.config_project_id}')
+    logging.info(f'Import: {absolute_import_path}')
+    logging.info(f'Import script args: {args_list}')
+    logging.info(f'Repo root directory: {repo_dir}')
+
+    cfg = _get_cloud_config()
```
ajaits, line 200: Would it be useful to test with a local config file to try other config options like script timeout, user script args before updating config for a job? Should such configs always be uploaded to GCS to test?

import-automation/executor/schedule_update_import.sh
jehangiramjad: Done!

import-automation/executor/README.md
```diff
@@ -47,13 +47,39 @@ Commons knowledge graph using the importer.
...
+Run `./schedule_update_import.sh -u <config_project_id> <path_to_import> <import_script_args>`
+```
+
+Run `./schedule_update_import.sh --help` for usage.
+
+`<config_project_id>` is the GCP project id where the config file is stored, e.g. `datcom-import-automation`.
+`<path_to_import>` is the path to the import (relative to the root directory of the `data` repo), with the name of the import provided with a colon, e.g. `scripts/us_usda/quickstats:UsdaAgSurvey`.
+`import_script_args` is one string (enclosed by quotes) representing all the command line args to be provided to the import script, e.g. `"--start_year=2023 --another_flag=val"`.
+
+Example invokation:
```
ajaits, line 70: s/invokation/invocation/

import-automation/executor/schedule_update_import.py
jehangiramjad: My thought is that there exist two ways to run it off of the latest master state of the repo:

1. You can run this same script from the master branch locally and it would be in the same state as running it off of the master state on the remote repo.
2. There is a workflow which allows these updates to be executed from Cloud Scheduler which will always download the latest repo and then execute.

The reason I wanted to keep this disabled on the local script is that our repo is ~500Mb and for every execution of this script to have to wait to download the repo first before executing is painful.

import-automation/executor/schedule_update_import.py
jehangiramjad: I can add that as a TODO where the user could provide a local config file to overwrite or add config params locally. Hopefully ok to do in a follow-up PR? 

Note that you are expected to provide the user script args locally as flag inputs to this script. So that won't be coming from GCS. However, the various auth params should only be hosted on GCS and I don't suspect those will change.

import-automation/executor/README.md
jehangiramjad: Done, thanks!

import-automation/executor/schedule_update_import.py
jehangiramjad: Added the TODO.

jehangiramjad: And here is the follow-up PR for this: https://github.com/datacommonsorg/data/pull/978

jehangiramjad: As a result of this, I have also removed the option to provide user script args as a command line flag. Instead, they can be set in the local config json file in the follow-up PR. Thanks for the suggestion!

ajaits (approved):

PR #978, opened 2024-02-04
[Import Automation] Support overriding GCS configs locally

jehangiramjad:
Follow-up from a discussion on the previous PR: https://github.com/datacommonsorg/data/pull/977#discussion_r1477317137


ajaits (approved):
Thanks Jehangir!

PR #986, opened 2024-02-21
Fix autorefresh manifest for floods script

ajaits:
Fix the import name and script in the floods manifest.json


jehangiramjad (approved):
Thanks!

PR #992, opened 2024-02-23
Disable gunicorn timeout for automation scripts

ajaits:
Disable timeout for gunicorn. 
The import executor has configurable timeouts for steps like repo download, user script that may exceed the 
gunicorn timeout of 1800 secs.


import-automation/executor/Dockerfile
```diff
@@ -27,4 +27,4 @@ RUN pip install -r /workspace/requirements.txt
 
 COPY app/. /workspace/app/
 
-CMD gunicorn --timeout 1800 --workers 5 -b :$PORT app.main:FLASK_APP
+CMD gunicorn --timeout 0 --workers 5 -b :$PORT app.main:FLASK_APP
```
jehangiramjad, line 30: question: setting this to 0 means no timeout?

jehangiramjad (approved):
Thanks

import-automation/executor/Dockerfile
ajaits: Yes, this disabled gunicorn timeout. The executor uses separate timeouts for each step.

PR #1021, opened 2024-06-05
Statvar Readme files for Brazil Import

HarishC727:


statvar_imports/brazil_visdata/FoodBasketDistribution/common_metadata.csv
```diff
@@ -0,0 +1,17 @@
+parameter,value
+url,
+description,Acquisition of Basic Baskets for families for Specific Population
+#place_type,
+#places_within,
+start_date,
```
ajaits, line 6: Let's add a reference to the places_csv
```
places_resolved_csv,../Brazil_Places_Resolved.csv
```

ajaits (approved):

statvar_imports/brazil_visdata/FoodBasketDistribution/common_metadata.csv
HarishC727: Can we keep it at both places - Metadata and the in the command line?
places_resolved_csv does not get recognised when placed inside the metadata but it works fine when passed as a flag in the command.

PR #1031, opened 2024-06-26
Contains the test data for india nfhs, place resolver csv file and readme

Vshnuz:
Contains the test data for india nfhs, place resolver csv file and readme file for the import.


Vshnuz:
There was a test issue, So I have re raised it

PR #1032, opened 2024-06-27
Test data for India along with the place resolved sheet and the Readme

Vshnuz:


statvar_imports/ndap/india_nfhs/test_data/sample_input/NFHS-State-pvmap.csv
ajaits, line 1: pls move pvmaps, metadata to the parent of test_data folder

PR #1082, opened 2024-09-30
PR  for EIA opendata Refresh

Harsha-chandaluri:


scripts/us_eia/opendata/process/common.py
```diff
@@ -21,7 +21,10 @@
 from sys import path
 
 # For import util.alpha2_to_dcid
-path.insert(1, '../../../../')
+# path.insert(1, '../../../../')
+path.insert(1, '/usr/local/google/home/chandaluri/data/')
```
shamimansari1988, line 25: You should not have a hard coded local path, please make general. Please modify the code as suggested.

scripts/us_eia/opendata/process/common.py
```diff
@@ -21,7 +21,10 @@
...
+path.insert(2,
+            '/usr/local/google/home/chandaluri/data/scripts/us_eia/opendata/')
```
shamimansari1988, line 27: You should not have a hard coded local path, please make general. Please modify the code as suggested.

PR #1090, opened 2024-10-12
Update email notification recipients

hqpho:
- Use datacommons-debug for successes and failures
- Use datacommons-alerts for failures
- Use category alias +imports for both addresses


beets (commented):
i think failure and success emails should both go to the same place. can we update failures to be sent to both places? it'll be much easier to find a complete record of all the runs in a single place.

hqpho:
> i think failure and success emails should both go to the same place. can we update failures to be sent to both places? it'll be much easier to find a complete record of all the runs in a single place.

Done!

beets (approved):
thanks for the update

PR #1103, opened 2024-10-30
Pep by srh data refresh for automation

Girish3320:


Girish3320:
additional files to be removed

PR #1115, opened 2024-11-13
Add dataset import differ utility

vish-cs:


tools/differ/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This utility generates a diff (point and series analysis) of two data files of the same dataset for import analysis.
+
+**Usage**
+```
+python differ.py --currentData=<filepath> --previousData=<filepath>
+```
+
+Parameter description
+currentDataFile: Location of the current MCF data file
```
ajaits, line 11: suffix `File` can be removed

tools/differ/README.md
```diff
@@ -0,0 +1,23 @@
...
+previousDataFile: Location of the previous MCF data file
+groupbyColumns: Columns to group data for diff analysis in the order var,place,time etc. Default value: “variableMeasured,observationAbout,observationDate”
+valueColumns: Columns with statvar value (unit etc.) for diff analysis. Default value:  "value,unit"
```
ajaits, line 14: Wondering if unit should be a group-by column.
For instance, if there are 2 SVObs for Temp in Celsius and Fahrenheit,
Prev:Max_Temp,PlaceDcid1,2024,80,F
Prev:Max_Temp,PlaceDcid1,2024,26.2,C

Cur:Max_Temp,PlaceDcid1,2024,27,C
Would this be compared to (Prev:Max_Temp,PlaceDcid1,2024,26.2,C) or (Prev:Max_Temp,PlaceDcid1,2024,80,F)?

tools/differ/differ.py
```diff
@@ -0,0 +1,167 @@
...
+from absl import flags
+from absl import logging
+import pandas as pd
+import helper
+import os
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('current_data', '', 'Path to the current MCF data.')
+flags.DEFINE_string('previous_data', '', 'Path to the previous MCF data.')
+flags.DEFINE_string('output_location', 'results', 'Path to the output data.')
```
ajaits, line 25: If this is expected to be a folder, pls add to the docstring
'Path to the output data folder.'

tools/differ/differ.py
```diff
@@ -0,0 +1,167 @@
...
+    self.place_column  =  self.groupby_columns[1]
+    self.time_column  =  self.groupby_columns[2]
+    self.diff_column = '_diff_result'
+
+  def __cleanup_data(self, df: pd.DataFrame):
+    for column in ['added', 'deleted', 'modified', 'same']:
+      df[column] = df[column] if column in df.columns else 0
+      df[column] = df[column].fillna(0).astype(int)
+
+  # Pro-rocesses two dataset files to identify changes.
```
ajaits, line 66: typo: Process

tools/differ/differ.py
```diff
@@ -0,0 +1,167 @@
...
+    return result
+
+  # Performs point diff analysis to identify data point changes.
+  def point_analysis(self, in_data: pd.DataFrame) -> (pd.DataFrame, pd.DataFrame):
+    column_list = [self.variable_column, self.place_column, self.time_column, self.diff_column]
+    result = in_data.loc[:, column_list]
+    # summary = summary.groupby([variable,'result'], observed=True, as_index=False).agg(['count']).reset_index()
+    result = result.groupby([self.variable_column, self.diff_column], observed=True, as_index=False)[[self.place_column, self.time_column]].agg(lambda x: x.tolist())
+    result['size'] = result.apply(lambda row:len(row[self.place_column]), axis=1)
+    result[self.place_column] = result.apply(lambda row: row[self.place_column][0:SAMPLE_COUNT], axis=1)
```
ajaits, line 95: If the places are sorted, can use `random.sample(list, size)` from the module random

tools/differ/differ.py
```diff
@@ -0,0 +1,167 @@
...
+    result[self.time_column] = result.apply(lambda row: row[self.time_column][0:SAMPLE_COUNT], axis=1)
```
ajaits, line 96: Can we keep the min/max and some samples in between?
```
  years = sorted(row[self.time_column])
  return years[0] + random.sample(years[1:-1] + years[-1]
```

tools/differ/differ.py
```diff
@@ -0,0 +1,167 @@
...
+    # result = result.groupby(
+    #   [self.variable_column, self.diff_column], observed=True, as_index=False).size()
+    summary = result.pivot(
+      index=self.variable_column, columns=self.diff_column, values='size')\
+      .reset_index().rename_axis(None, axis=1)
+    self.__cleanup_data(summary)
```
ajaits, line 102: single underscore for private functions?
`_cleanup_data()`

tools/differ/differ.py
vish-cs: Fixed

tools/differ/differ.py
vish-cs: Done

tools/differ/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Output generated is of the form below showing counts of differences for each variable.
+Detailed diff output is written to a file for further analysis.
+
+variableMeasured   added  deleted  modified  same  total
+0   dcid:var1       1      0       0          0     1
+1   dcid:var2       0      2       1          1     4
+2   dcid:var3       0      0       1          0     1
+3   dcid:var4       0      2       0          0     2
```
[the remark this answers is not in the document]
vish-cs: Done

tools/differ/differ.py
vish-cs: Done

tools/differ/differ.py
vish-cs: Done

tools/differ/README.md
vish-cs: I am not sure. If we add it as a group by column, it will treat a change in the unit as a new series. If we add it to value columns, it will treat it as a modified data point. I think the latter makes more sense.

ajaits: If an import has same data in multiple units, if the unit in value can compare the right units, then this is fine:
say: previous has 
Max_Temp,PlaceDcid1,2024,80,F
Max_Temp,PlaceDcid1,2024,26.2,C

current has:
Max_Temp,PlaceDcid1,2024,26.21,C
Max_Temp,PlaceDcid1,2024,80,F

Only the first in current should modified. It shouldn;t compare F with C due to the change in output ordering.
If the current version does that we can go ahead.

tools/differ/differ.py
```diff
@@ -0,0 +1,251 @@
...
+    result.sort_values(by=[differ.diff_column, differ.variable_column],
+                       inplace=True)
+    print(summary.head(10))
+    print(result.head(10))
+    helper.write_data(summary, FLAGS.output_location,
+                      'series-analysis-summary.csv')
+    helper.write_data(result, FLAGS.output_location,
+                      'series-analysis-results.csv')
+
+    logging.info('Differ output written to folder: %s', FLAGS.output_location)
```
ajaits, line 247: Can this be moved into a function that other modules can directly invoke?

Since it is in the same repo, the import-executor could make a function call instead of a sub-process

tools/differ/helper.py
```diff
@@ -0,0 +1,88 @@
...
+
+def load_mcf_file(file: str) -> pd.DataFrame:
+    """ Reads an MCF text file and returns it as a dataframe."""
+    mcf_file = open(file, 'r', encoding='utf-8')
+    mcf_contents = mcf_file.read()
+    mcf_file.close()
+    # nodes separated by a blank line
+    mcf_nodes_text = mcf_contents.split('\n\n')
+    # lines seprated as property: constraint
+    mcf_line = re.compile(r'^(\w+): (.*)$')
```
ajaits, line 17: there can be spaces around `:`
for example: `Node : dcid:TestNode` is valid.

tools/differ/helper.py
```diff
@@ -0,0 +1,88 @@
...
+    mcf_nodes = []
+    for node in mcf_nodes_text:
+        current_mcf_node = {}
+        for line in node.split('\n'):
+            parsed_line = mcf_line.match(line)
+            if parsed_line is not None:
+                current_mcf_node[parsed_line.group(1)] = parsed_line.group(2)
+        if current_mcf_node and current_mcf_node[
+                'typeOf'] == 'dcid:StatVarObservation':
+            mcf_nodes.append(current_mcf_node)
```
ajaits, line 27: Can we log some warning for non SVObs nodes ignored?

tools/differ/helper.py
```diff
@@ -0,0 +1,88 @@
...
+
+def load_mcf_files(path: str) -> pd.DataFrame:
+    """ Loads all sharded mcf files in the given directory and 
+    returns a single combined dataframe."""
+    df = pd.DataFrame()
+    filenames = glob.glob(path + '.mcf')
+    for filename in filenames:
+        df2 = load_mcf_file(filename)
+        # Merge data frames, expects same headers
+        df = pd.concat([df, df2])
```
ajaits, line 40: collecting all dfs into a list and a single concat at the end may save some copies.

tools/differ/helper.py
```diff
@@ -0,0 +1,88 @@
...
+    Returns:
+      path to the output file/folder
+    """
+
+    client = Client()
+    bucket = client.get_bucket(uri.split('/')[2])
+    if uri.endswith('*'):
+        blobs = client.list_blobs(bucket)
+        for blob in blobs:
+            path = os.path.join(os.getcwd(), tmp_dir, blob.name)
```
ajaits, line 80: can blob.download_to_filename() take relative paths? 
if so, can we skip getcwd()? that way if the tmp_dir is an absolute path, file is written there.

tools/differ/helper.py
```diff
@@ -0,0 +1,88 @@
...
+            blob.download_to_filename(path)
+        return os.path.join(os.getcwd(), tmp_dir, '*')
+    else:
+        file_name = uri.split('/')[3]
+        blob = bucket.get_blob(file_name)
+        path = os.path.join(os.getcwd(), tmp_dir, file_name)
```
ajaits, line 86: do we need os.path.basename(file_name) to get last part of the file ?

o/w we may need to create the sub-directories unless download_to_filenmae() creates missing folders too.

tools/differ/test/result1.csv
ajaits, line 1: Can we add sample summary files too? even if we don't use it in tests?

ajaits (approved):

tools/differ/differ.py
vish-cs: Moved the logic to helper function run_differ

tools/differ/helper.py
vish-cs: Done

tools/differ/helper.py
vish-cs: Added a log message

tools/differ/helper.py
vish-cs: I have removed getcwd part

tools/differ/test/result1.csv
vish-cs: I have added them as part of the checked in code

tools/differ/helper.py
vish-cs: Updated the expression to  mcf_line = re.compile(r'^(\w+)\s*:(.*)$')

tools/differ/README.md
vish-cs: We concatenate value+unit to create the target value which is then compared. Having two different value points for the same sv_ob is not supported in the implementation.

tools/differ/helper.py
vish-cs: Changed the implementation to replace '/' with '_' in the file name

tools/differ/README.md
vish-cs: Added unit to groupby_columns as per offline discussion

tools/differ/test/results/series-analysis-results.csv
```diff
@@ -0,0 +1,6 @@
+variableMeasured,_diff_result,observationAbout,size
```
ajaits, line 1: Can we drop the leading '_' for `diff_result`?

ajaits (approved):

tools/differ/test/results/series-analysis-results.csv
vish-cs: done

PR #1132, opened 2024-12-09
Update utils for statvar processor

ajaits:
Update utils used by statvar processor


util/config_map.py
```diff
@@ -55,28 +55,32 @@
...
-                 config_dict: dict = None,
-                 filename: str = None,
-                 config_string: str = None):
-        '''Create a Config Map object.
-        Args:
-          config_dict: dictionary with key:values to be loaded into the config map.
-          filename: override the dictionary with key:values from the file.
-          config_string: string of dictionary parameters to override key:values.
-        '''
+    """Class to store config mapping of named parameters to values as a dictionary."""
```
rohitkumarbhagat, line 69: Looks like a formatting change (indentation + ''' -> """). Is this per Google Style guide recommendation?

util/download_util.py
```diff
@@ -166,14 +173,16 @@ def request_url(url: str,
...
-                    logging.debug(
-                        f'Retrying URL {url} after {retry_secs} secs ...')
-                    time.sleep(retry_secs)
+                    requests.exceptions.ConnectionError, urllib.error.URLError,
+                    urllib.error.HTTPError) as e:
+                logging.debug(f'Got exception {e} for {url}, {params}')
+
+            # retry in case of errors
+            if attempt >= retries:
+                raise urllib.error.URLError
```
rohitkumarbhagat, line 182: Should we not just bubble up the original error instead using a fixed error. This would help debug without enabling debug logs.

util/file_util.py
```diff
@@ -428,9 +432,12 @@ def file_get_name(file_path: str,
   Returns:
     file name combined from path, suffix and extension.
   """
-    # Create the file directory if it doesn't exist.
+    if not file_path:
+        return ''
```
rohitkumarbhagat, line 436: Should we return error for blank file paths rather than empty? It would be easy to debug if empty path is the culprit

util/file_util.py
```diff
@@ -540,7 +548,7 @@ def file_load_csv_dict(
     config: dictionary of aggregation settings in case there are multiple rows
       with the same key. refer to aggregation_util.aggregate_dict for config
       settings.
-
+    key_index: if True, each rows is loaded with a unique key for row index.
```
rohitkumarbhagat, line 551: Per implementation, `key_index` gets preference over `key_column`. Should we disallow key_column when key_index = True?

util/file_util.py
```diff
@@ -629,6 +643,9 @@ def file_write_csv_dict(py_dict: dict,
       is used as the key's column name. If no columns are specified for values,
       column names are picked from each entry's value if the value is a dict.
       Else the value is written as column name 'value'.
+    key_column_name: name of the column used as key.
```
rohitkumarbhagat, line 646: For backward compatibility, should the default value of key_column_name not be None instead of 'key' . Previous implementation picked first column as key column. Same behavior would be preserved using None as key_column_name.

That being said, will it be more intuitive to use None to skip key and something else (Enum/constant) to indicate first column as key?

util/file_util.py
```diff
@@ -629,6 +643,9 @@ def file_write_csv_dict(py_dict: dict,
...
+      if None, the first column is used as key.
```
rohitkumarbhagat, line 647: Will it be more intuitive to use None to indicate key is ignored?

util/file_util.py
```diff
@@ -811,8 +833,17 @@ def file_open_google_spreadsheet(url: str) -> gspread.spreadsheet.Spreadsheet:
     google spreadsheet object for the given url
   """
     # Get a handle for the whole spreadsheet
-    gs = _file_get_gspread_client().open_by_url(url)
-    return gs
+    while retries >= 0:
```
rohitkumarbhagat, line 836: optional: should we use retry module  https://pypi.org/project/retry/?

util/file_util.py
```diff
@@ -963,6 +997,121 @@ def file_copy_to_spreadsheet(filename: str,
     return ws.url
 
 
+def file_get_sample_bytes(file: str, size: int = 4096) -> bytes:
```
rohitkumarbhagat, line 1000: nit: size -> byte_count?

util/file_util.py
```diff
@@ -963,6 +997,121 @@ def file_copy_to_spreadsheet(filename: str,
...
+    # Get the char with the same count across rows.
+    for c in char_counts.keys():
+        c_counts = char_counts[c]
+        if c_counts:
+            c_min = min(c_counts)
+            c_med = np.median(c_counts)
+            if c_min > 0 and c_min == c_med:
+                result['delimiter'] = c
+                break
+    logging.level_debug() and logging.debug(
```
rohitkumarbhagat, line 1110: Logging level check could be skipped if we use logging.debug with arguments like 
logging.debug("Got options for file: %s, result = %s", file, result)

rohitkumarbhagat (commented):
Left few comments. Please take a look

util/config_map.py
ajaits: This was an automated change running `pylint` when yapf was deprecated in google3.

`yapf --style google` allows both and I have switched back to using it for more recent code.

util/download_util.py
ajaits: We would like to retry on errors like timeout, server error etc which are also exceptions hence the original error is consumed with a log message and retry attempted.

util/file_util.py
ajaits: by 'return error' do you mean `return None` or throw an exception? 

Throwing an exception complicates caller who has to catch the error rather than check for returned value.

the `if` check for return type works with both  `None and ``. Is the convention to return None rather than empty? The code may not be consistent in this as it uses a simple if check.

util/file_util.py
ajaits: Added an error log if key_column and key_index are both set.

util/config_map.py
rohitkumarbhagat: lgtm

util/file_util.py
rohitkumarbhagat: I meant throwing an exception. Usually, for precondition/validation failures, ValueError is thrown by many python libraries. Indicating error via some code( returning blank) vs exceptions have their pros and cons. I do not see there being a convention. 

Betwee n`` and None, None seems more intuitive to me.

util/file_util.py
rohitkumarbhagat: Sorry, I am not able to locate error log. IMO, error logs are easy to miss. Throwing precondition error (ValueError) might be a tighter enforcement.

util/file_util.py
ajaits: Good suggestion. Done

util/file_util.py
ajaits: Done

util/file_util.py
ajaits: Done

util/file_util.py
ajaits: Done. will try the AI tool to auto-correct these.

util/file_util.py
ajaits: Changed to None. The callers in rest of the code may not handle exceptions, can update this to more consistent error in future.

util/file_util.py
```diff
@@ -540,18 +549,26 @@ def file_load_csv_dict(
...
   Returns:
     dictionary of {key:value} loaded from the CSV file.
   """
     csv_dict = {}
     input_files = file_get_matching(filename)
     logging.debug(f'Loading dict from csv files: {input_files}')
+    if key_column and key_index:
+        raise ValueError(
+            f'Both Key_column: {key_column} and key_index set for {filename}')
+
```
rohitkumarbhagat, line 563: Thanks for adding the validation.

rohitkumarbhagat (approved):

PR #1139, opened 2024-12-11
Add basic validation logic

vish-cs:


tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2020 Google LLC
```
ajaits, line 1: change year to 2024?

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+  ('ADDED_COUNT', 3),
+  ('DELETED_COUNT', 4),
+  ('LATEST_DATA', 5),
+])
+
+class ImportValidation:
+  '''
+  Import validation
+  '''
+  def __init__(self, config_file: str):
```
ajaits, line 46: can we pass in the csv filename and load it once instead of in each validate function?

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+    }
+    with open(config_file, encoding='utf-8') as fd:
+      validation_config = json.load(fd)
+      for config in validation_config:
+        self.run_validation(config)
+
+  def _latest_data_validation(self, config: dict):
+    logging.info('Not yet implemented')
+
+  def _deleted_count_validation(self, config: dict):
```
ajaits, line 62: Can we have a one line description of the expected count?

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+    df = pd.read_csv(os.path.join(
+      FLAGS.differ_output_location, POINT_ANALAYSIS_FILENAME))
+    if df['deleted'].sum() > config['threshold']:
```
ajaits, line 65: Is the threshold common for all checks, deleted, modified, etc or is this specific to each check?

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+
+  def _unmodified_count_validation(self, config: dict):
+    df = pd.read_csv(os.path.join(
+      FLAGS.differ_output_location, POINT_ANALAYSIS_FILENAME))
+    if df['same'].nunique() > 1:
+      raise AssertionError(f'Validation failed: {config["validation"]}')
+
+  def run_validation(self, config):
+    try:
+      self.validation_map[Validation[config['validation']]](config)
```
ajaits, line 88: Can the config run multiple validations?
If so, can we collect errors and raise exception with all errors listed?
this makes it easier to troubleshoot when there are multiple validation check failures.

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+  ('MODIFIED_COUNT', 1),
+  ('UNMODIFIED_COUNT', 2),
+  ('ADDED_COUNT', 3),
+  ('DELETED_COUNT', 4),
+  ('LATEST_DATA', 5),
+])
+
+class ImportValidation:
+  '''
+  Import validation
```
ajaits, line 44: Can you pls add a comment on usage with a sample config?

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+      validation_config = json.load(fd)
+      for config in validation_config:
+        self.run_validation(config)
+
+  def _latest_data_validation(self, config: dict):
+    logging.info('Not yet implemented')
+
+  def _deleted_count_validation(self, config: dict):
+    df = pd.read_csv(os.path.join(
+      FLAGS.differ_output_location, POINT_ANALAYSIS_FILENAME))
```
ajaits, line 64: let's pass in values rather than use command line flags directly within the class as it makes it easier to test.

tools/validation/validation.py
```diff
@@ -0,0 +1,97 @@
...
+  def _unmodified_count_validation(self, config: dict):
+    df = pd.read_csv(os.path.join(
+      FLAGS.differ_output_location, POINT_ANALAYSIS_FILENAME))
+    if df['same'].nunique() > 1:
+      raise AssertionError(f'Validation failed: {config["validation"]}')
+
+  def run_validation(self, config):
+    try:
+      self.validation_map[Validation[config['validation']]](config)
+      logging.info('Validation passed: %s', config['validation'])
```
ajaits, line 89: Of there are no changes at all should we fail the autorefresh?

tools/validation/validation.py
vish-cs: done

tools/validation/validation.py
vish-cs: Each check can define its own threshold in the config.

tools/validation/validation.py
vish-cs: Good point. Done

tools/validation/validation.py
vish-cs: Done

tools/validation/validation.py
vish-cs: I am not sure. That could be import specific I think. We should make that as a separate validation if needed.

tools/validation/validation.py
vish-cs: done

tools/validation/validation.py
vish-cs: Done

tools/validation/validation.py
vish-cs: We currently run all the validations. I have updated the logic to log the status for each validation along with error message.

tools/import_validation/import_validation.py
```diff
@@ -0,0 +1,136 @@
...
+        with open(config_file, encoding='utf-8') as fd:
+            self.validation_config = json.load(fd)
+
+    def _latest_data_validation(self, config: dict):
+        logging.info('Not yet implemented')
+
+    # Checks if the number of deleted data points are below a threshold.
+    def _deleted_count_validation(self, config: dict):
+        if self.differ_results['deleted'].sum() > config['threshold']:
+            raise AssertionError(f'Validation failed: {config["validation"]}')
```
ajaits, line 90: Can we add more context, such as `deletions <count> > <threshold>`

tools/import_validation/import_validation.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Checks if number of modified points for each stat var are same.
+    def _modified_count_validation(self, config: dict):
+        if self.differ_results['modified'].nunique() > 1:
+            raise AssertionError(f'Validation failed: {config["validation"]}')
```
ajaits, line 95: Can we list some sample statvars without modifications?

ajaits (approved):

tools/import_validation/import_validation.py
vish-cs: I will do this in a follow up PR

tools/import_validation/validation_output.csv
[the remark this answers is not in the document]
vish-cs: Done

tools/import_validation/import_validation.py
vish-cs: I will do this in a follow up PR

PR #1141, opened 2024-12-16
Invoke validations from import executor

vish-cs:


import-automation/executor/app/configs.py
```diff
@@ -111,6 +111,12 @@ class ExecutorConfig:
     user_script_args: List[str] = ()
     # Environment variables for the user script
     user_script_env: dict = None
+    # Skip uploading the data to GCS (for local testing).
+    skip_gcs_upload: bool = True
+    # Invoke validations before upload.
+    invoke_import_validation: bool = True
+    # Import validation config file.
+    validation_config_file: str = 'tools/validation/config.json'
```
ajaits, line 119: can we call this validation_config.json? 
There are a lot of config.json files in the repo.

import-automation/executor/app/executor/import_executor.py
```diff
@@ -318,6 +319,100 @@ def _import_one(
...
+        """
+        import_inputs = import_spec.get('import_inputs', [])
+        for import_input in import_inputs:
+            mcf_path = import_input['node_mcf']
+            current_data_path = os.path.join(absolute_import_dir, mcf_path)
+            previous_data_path = os.path.join(absolute_import_dir,
+                                              mcf_path) + '.old'
+            differ_results_path = os.path.join(absolute_import_dir, 'results')
+            config_file_path = os.path.join(absolute_import_dir,
+                                            self.config.validation_config_file)
```
ajaits, line 337: Can this be an import specific validation config declared in the manifest.json and default to config.validation_config_file if not declared?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -318,6 +319,100 @@ def _import_one(
...
+            # Invoke data differ script.
+            differ_script_path = os.path.join(repo_dir, 'tools', 'differ',
+                                              'differ.py')
+            differ_script_args: List[str] = ('--current_data=' +
+                                             current_data_path,
+                                             '--previous_data=' +
+                                             previous_data_path,
+                                             '--output_location=' +
+                                             differ_results_path)
+            process = _run_user_script(
```
ajaits, line 357: Can we call the differ module directly instead of a process?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -351,36 +446,23 @@ def _import_one_helper(
...
-                    cloud_run_simple_import.cloud_run_simple_import_job(
-                        import_spec=import_spec,
-                        config_file=script_path,
-                        env=self.config.user_script_env,
-                        version=version,
-                        image=import_spec.get('image'),
-                    )
-                else:
-                    # Run import script locally.
-                    script_interpreter = _get_script_interpreter(
```
ajaits, line 370: Pls retain this as the interpreter path changes for imports using shell scripts

import-automation/executor/app/configs.py
```diff
@@ -111,6 +111,12 @@ class ExecutorConfig:
     user_script_args: List[str] = ()
     # Environment variables for the user script
     user_script_env: dict = None
+    # Skip uploading the data to GCS (for local testing).
+    skip_gcs_upload: bool = True
```
ajaits, line 115: Should we turn this to False in the config_overrides.json?

import-automation/executor/app/configs.py
```diff
@@ -111,6 +111,12 @@ class ExecutorConfig:
...
+    # Invoke validations before upload.
+    invoke_import_validation: bool = True
```
ajaits, line 117: Let's keep this as False until we add call to dc-import so the previous version of MCF is stored in GCS

import-automation/executor/app/configs.py
vish-cs: I have removed this flag from this PR

import-automation/executor/app/configs.py
vish-cs: Done

import-automation/executor/app/configs.py
vish-cs: Done

import-automation/executor/app/executor/import_executor.py
vish-cs: Yes I have added a conditional assignment to check for it in the import spec first, else use the default value from the config

import-automation/executor/app/executor/import_executor.py
vish-cs: This logic is not removed..it's just refactored out into a separate function

import-automation/executor/app/executor/import_executor.py
```diff
@@ -330,57 +340,49 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
...
             bucket = storage.Client(self.config.gcs_project_id).bucket(
                 self.config.storage_prod_bucket_name)
             folder = relative_import_dir + '/' + import_spec['import_name'] + '/'
             blob = bucket.blob(folder + 'latest_version.txt')
+            if not blob:
+                logging.error(
+                    "Not able to download latest version from GCS, skipping validation."
+                )
+                return
             blob = bucket.blob(folder + blob.download_as_text() + '/' +
```
ajaits, line 364: Before reassigning blob, can we save the old version from blob.download_as_text() into a local variable and add that to the error messages?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -330,57 +340,49 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
...
+            logging.info(f'Validation config file: {config_file_path}')
 
             # Download previous import data.
             bucket = storage.Client(self.config.gcs_project_id).bucket(
                 self.config.storage_prod_bucket_name)
             folder = relative_import_dir + '/' + import_spec['import_name'] + '/'
             blob = bucket.blob(folder + 'latest_version.txt')
+            if not blob:
+                logging.error(
+                    "Not able to download latest version from GCS, skipping validation."
```
ajaits, line 361: Can we add the folder name looked up in the message too?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -330,57 +340,49 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
...
-                interpreter_path=interpreter_path,
-                script_path=validation_script_path,
-                timeout=self.config.user_script_timeout,
-                args=validation_script_args,
-                cwd=absolute_import_dir,
-                env=self.config.user_script_env,
-            )
-            _log_process(process=process)
-            process.check_returncode()
+            if not blob:
```
ajaits, line 366: If mcf_path is empty, can we skip this or add a todo to generate mcf_path from dc-import output.

import-automation/executor/app/executor/import_executor.py
ajaits: I don't see any call to _get_script_interpreter() in the new code.
Pls nod ethat there are two places where interpreter_path is computed:
1. _create_venv() generates an intepreter path for python scripts based on local pip install.
2. To invoke the import['user_script'], if it is a non-python script such as shell or jar, the _get_script_interpreter() generates a different interpreter. 
  This was added recently to support shell scripts.
  _invoke_import_job() is currently calling all scripts with interpreter_from _create_venv() that will only work for py and imports with break shell scripts.

tools/import_differ/import_differ.py
```diff
@@ -23,6 +23,10 @@
 
 import differ_utils
 
+SAMPLE_COUNT = 3
+GROUPBY_COLUMNS = 'variableMeasured,observationAbout,observationDate,measurementMethod,unit'
```
ajaits, line 27: can also add `observationPeriod`

vish-cs: Done

import-automation/executor/app/executor/import_executor.py
vish-cs: Done

import-automation/executor/app/executor/import_executor.py
vish-cs: Done

import-automation/executor/app/executor/import_executor.py
vish-cs: Added a check to skip validations if node_mcf is empty in manifest

ajaits (approved):

import-automation/executor/app/executor/import_executor.py
```diff
@@ -403,6 +407,8 @@ def _invoke_import_job(self, absolute_import_dir: str, import_spec: dict,
                 )
             else:
                 # Run import script locally.
+                script_interpreter = _get_script_interpreter(
+                    script_path, interpreter_path)
                 process = _run_user_script(
                     interpreter_path=interpreter_path,
```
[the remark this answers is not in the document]
vish-cs: Thanks for catching

PR #1142, opened 2024-12-16
NCES_PublicSchoolStats_import_refresh

Girish3320:


scripts/us_nces/README_download.md
```diff
@@ -0,0 +1,55 @@
+# Downloading input files for US: National Center for Education Statistics
+
```
shamimansari1988, line 2: Could you please keep only one README.md file and mention download steps there which you wrote in "README_Download.md" file.

Girish3320:
We have common PR for public, district and private.

PR #1173, opened 2025-01-04
Rename differ utility files

vish-cs:


tools/import_differ/file_utils.py
ajaits, line 1: there is a similar filename: https://github.com/datacommonsorg/data/blob/master/util/file_util.py. 
Perhaps name as import_differ_utils?

ajaits (approved):

tools/import_differ/file_utils.py
vish-cs: Renamed to differ_utils

PR #1186, opened 2025-01-16
Adds logs and labels in auto-import jobs to help create log based metrics.

rohitkumarbhagat:


import-automation/executor/app/executor/cloud_run.py
```diff
@@ -61,11 +61,13 @@ def create_or_update_cloud_run_job(project_id: str, location: str, job_id: str,
 
     res = run_v2.types.ResourceRequirements(limits=resources)
     container = run_v2.Container(image=image, env=env, resources=res, args=args)
+    # Labels allow filtering of automated import cloud run jobs, used in log-based metrics.
     exe_template = run_v2.ExecutionTemplate(
-        template=run_v2.TaskTemplate(containers=[container],
-                                     max_retries=2,
-                                     timeout=duration_pb2.Duration(
-                                         seconds=timeout)))
+        labels={"datacommons_cloud_run_job_type": "auto-import_job"},
```
vish-cs, line 66: auto-import_job: can we change to auto_import_job?

import-automation/executor/main.py
```diff
@@ -35,6 +46,20 @@ def scheduled_updates(absolute_import_name: str, import_config: str):
         local_repo_dir=config.local_repo_dir)
     result = executor.execute_imports_on_update(absolute_import_name)
     logging.info(result)
+    elapsed_time_secs = int(time.time() - start_time)
+    message = (f"Cloud Run Job [{CLOUD_RUN_JOB_NAME}] completed with status= "
```
vish-cs, line 50: Don't we get this info from Cloud Run directly (dashboard shows this)?

import-automation/executor/main.py
```diff
@@ -14,8 +16,17 @@
 flags.DEFINE_string('import_name', '', 'Absoluate import name.')
 flags.DEFINE_string('import_config', '', 'Import executor configuration.')
 
+CLOUD_RUN_JOB_NAME = os.getenv("CLOUD_RUN_JOB")
```
vish-cs, line 19: where do we set this env. var?

rohitkumarbhagat: This is set in contetx by cloudrun ref: https://cloud.google.com/run/docs/container-contract#jobs-env-vars

import-automation/executor/main.py
rohitkumarbhagat: Cloud run does not capture metrics for latency. The message here is used in structured json log (which would be filtered for log-based metric). The message is shown as log message by log explorer.

import-automation/executor/app/executor/cloud_run.py
rohitkumarbhagat: Thanks for catching this.

import-automation/executor/main.py
vish-cs: I meant to ask..where are we setting this env. var in the code?

import-automation/executor/main.py
vish-cs: The dashboard does show total run time for the job. Is that not what we need?

import-automation/executor/main.py
rohitkumarbhagat: IIUC, Cloud run execution automatically sets this environment variable before starting the job. Users do not have to set this explicitly.

vish-cs (approved):

import-automation/executor/main.py
rohitkumarbhagat: Discussed offline. Cloud Run does maintain latency numbers which could be fetched over api. It's not exposed as metric yet. In current implementation, we would log and create log-based metric for the same.

ajaits (approved):

PR #1192, opened 2025-01-21
Add git commit id as docker image tag

vish-cs:


import-automation/executor/cloudbuild.yaml
```diff
@@ -7,12 +7,12 @@
 steps:
   # Docker Build
   - name: 'gcr.io/cloud-builders/docker'
-    args: ['build', '-t', '${_DOCKER_IMAGE}:latest', '.']
+    args: ['build', '-t', '${_DOCKER_IMAGE}:{COMMIT_SHA}', '.']
```
rohitkumarbhagat, line 10: While adding `stable` tag, we are using `latest` as source tag. Can that fail, as now we are not tagging `latest`?

rohitkumarbhagat (changes requested):

import-automation/executor/cloudbuild.yaml
vish-cs: Fixed

import-automation/executor/cloudbuild.yaml
```diff
@@ -33,4 +33,4 @@ steps:
     - '-c'
     - |
       docker tag ${_DOCKER_IMAGE}:latest ${_DOCKER_IMAGE}:stable \
```
rohitkumarbhagat, line 35: Do we need to change the latest tag here too?

import-automation/executor/cloudbuild.yaml
rohitkumarbhagat: Can we also preserve both  latest and commit tags? latest tag can be helpful to locate latest image in artifact regisrty.

rohitkumarbhagat (changes requested):

import-automation/executor/cloudbuild.yaml
vish-cs: ok it seems the push did not pick up some changes from my local. Thanks for catching.

import-automation/executor/cloudbuild.yaml
rohitkumarbhagat: Thanks Vishal.

Optional: Can we also preserve both latest and commit tags? latest tag can be helpful to locate latest image in artifact registry.

vish-cs: Updated the PR to also retain the latest tag

rohitkumarbhagat: Sorry, but I can't locate this change. Has the PR missed your local changes?

rohitkumarbhagat (approved):

PR #1199, opened 2025-01-24
Clean up Python requirements

vish-cs:


requirements_all.txt
```diff
@@ -1,46 +1,17 @@
 # Requirements for all Python code in this repo, except for import-automation
```
ajaits, line 1: Can we add some instructions on where a new module should be added?
if the script using the module 
- may be used in import automation, add it to  import-automation/executor/requirements.txt
- is needed for running tests run locally, add here
- is only needed for locally executing the script without any tests, add to the requirements.txt file within the script folder

ajaits (approved):

rohitkumarbhagat (approved):

requirements_all.txt
vish-cs: Does this make sense?

Add new modules to: 
- import-automation/executor/requirements.txt: if the script is used in import automation.
- <script-folder>/requirements.txt: the module is only needed for that script.
- requirements_all.txt (here): anything not related to import automation.

requirements_all.txt
```diff
@@ -1,46 +1,17 @@
...
 
-absl-py
-arcgis2geojson
+-r import-automation/executor/requirements.txt
```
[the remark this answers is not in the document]
vish-cs: Ideally we should avoid this scenario. The one from the automation file be the default one which will get overwritten.

PR #1201, opened 2025-01-26
Fix Python requirement logic

vish-cs:


import-automation/executor/app/executor/import_executor.py
```diff
@@ -445,8 +448,10 @@ def _import_one_helper(
         with tempfile.TemporaryDirectory() as tmpdir:
             requirements_path = os.path.join(absolute_import_dir,
                                              self.config.requirements_filename)
+            central_requirements_path = os.path.join(
+                os.getcwd(), self.config.requirements_filename)
```
ajaits, line 452: If this is the data root, can we use the repo_dir?
getcwd() seems error prone as it assumes the executor script is run from the repo root.

ajaits (approved):

import-automation/executor/app/executor/import_executor.py
vish-cs: This is not repo_dir. This is import-automation/executor folder.

vish-cs: Updated to include full path

PR #1209, opened 2025-01-29
Added log_util.py for creating structured logs for log-based metric.

rohitkumarbhagat:


ajaits (approved):
Thanks Rohit.

As we add more metrics for logging, please add some guidelines into util/log_util.py  for
1. what to add a new property:value as a metric:label vs what to add to message
  such as it's impact to dashboards and monitoring configs.
2. can metric_labels have values as int, string, lists, etc
3. is there a way to get aggregations in the monitoring dashboard (sum of log messages of a certain type+label), something like a Counter metric.

rohitkumarbhagat:
Thanks Ajai for the review. Addressed following review comments. Please review.

> As we add more metrics for logging, please add some guidelines into util/log_util.py for
> 
> 1. what to add a new property:value as a metric:label vs what to add to message
>    such as it's impact to dashboards and monitoring configs.
Added
> 2. can metric_labels have values as int, string, lists, etc
Added

Added some guidelines on creating log-based metrics.

> 3. is there a way to get aggregations in the monitoring dashboard (sum of log messages of a certain type+label), something like a Counter metric.

Yes, counter metrics are supported. Added a note that metric labels might not be required in counter metric.

ajaits (approved):
Thanks Rohit!

PR #1224, opened 2025-02-03
Wildcard support for import differ

vish-cs:


tools/import_differ/differ_utils.py
```diff
@@ -55,36 +56,38 @@ def write_data(df: pd.DataFrame, path: str, file: str):
...
+      path: local or gcs path (single file or wildcard format)
       tmp_dir: destination folder
     Returns:
       dataframe with the input data
     """
     if path.startswith('gs://'):
         path = get_gcs_data(path, tmp_dir)
 
-    if path.endswith('*'):
+    if '*' in path:
```
ajaits, line 67: can we always do load_mcf_files() so it automatically eorks for other glob() patterns like ?

tools/import_differ/import_differ.py
```diff
@@ -218,13 +218,15 @@ def series_analysis(self,
...
 
     def run_differ(self):
-        if not os.path.exists(self.output_location):
-            os.makedirs(self.output_location)
         logging.info('Loading data...')
-        current_df = differ_utils.load_data(self.current_data,
-                                            self.output_location)
-        previous_df = differ_utils.load_data(self.previous_data,
-                                             self.output_location)
+        current_dir = os.path.join(self.output_location, 'current')
```
ajaits, line 222: Is current data expected to be in a local folder named 'current'? Autorefresh scripts may have their own folder heirarchy.

vish-cs: This directory is created by the script for storing temp data. We pass the input data folder name as an argument so the autorefresh script can use any path.

tools/import_differ/differ_utils.py
vish-cs: Done

ajaits (approved):
Thanks Vishal!

PR #1243, opened 2025-02-17
Dataflow based import differ

vish-cs:


tools/import_differ/differ_utils.py
```diff
@@ -46,24 +48,83 @@ def load_mcf_files(path: str) -> pd.DataFrame:
...
+        for filepath in glob.iglob(src):
+            filename = os.path.basename(filepath)
+            logging.info('Uploading %s to %s', filename, dest)
+            blobname = dest[len('gs://' + bucket_name + '/'):] + '/' + filename
+            blob = bucket.blob(blobname)
+            blob.upload_from_filename(filepath)
+    else:
+        os.makedirs(dest, exist_ok=True)
+        for filepath in glob.iglob(src):
+            shutil.copyfile(filepath, dest + '/' + os.path.basename(filepath))
```
ajaits, line 127: can do `os.path.join(dest, os.path.basename(filepath))` instead of adding '/'

tools/import_differ/import_differ.py
```diff
@@ -91,72 +112,88 @@ def __init__(self,
...
-          intermediate merged data for analysis
+        Runs dataflow job to process two datasets to identify changes.
         """
-        cur_df_columns = current_df.columns.values.tolist()
-        self.groupby_columns = [
-            i for i in self.groupby_columns if i in cur_df_columns
-        ]
-        self.value_columns = [
-            i for i in self.value_columns if i in cur_df_columns
+        if self.runner_mode == 'local':
```
ajaits, line 132: Wondering if we should retain the old python differ mode too for users who can't use java.

ajaits (approved):

tools/import_differ/differ_utils.py
vish-cs: Done

tools/import_differ/import_differ.py
vish-cs: The formats have diverged at this point. Also, the dataflow PR is already under review. I have uploaded a jar for anyone to use.

PR #1249, opened 2025-02-21
Eurostat_health_6_imports

SudhishaK:
Download urls are updated.


scripts/eurostat/health_determinants/common/euro_stat.py
```diff
@@ -27,7 +27,7 @@
 import pandas as pd
 import numpy as np
 from absl import flags, logging
-
+import inspect
```
krishnaswamypradeep, line 30: Please remove this module if it's only used for debugging purposes.

krishnaswamypradeep (approved):

PR #1256, opened 2025-02-27
UNEnergy

dhhhanushree:
UNEnergy_data_set_refresh


scripts/un/energy/process.py
```diff
@@ -479,6 +475,18 @@ def _process_row(data_row: dict, sv_map: dict, row_map: dict, sv_obs: dict,
     counters['output_csv_rows'] += 1
 
 
+def rename_the_mapping_value(sv_id, mapping_names):
```
krishnaswamypradeep, line 478: Concise code suggested by Gemini CLI:
```
def rename_the_mapping_value(sv_id, mapping_names):
    # Renaming the mapping value.
    if sv_id is None:
        return None
    key_to_check = sv_id.split(":", 1)[1]
    return sv_id.replace(key_to_check, mapping_names.get(key_to_check, key_to_check))
```

scripts/un/energy/download.py
```diff
@@ -64,12 +65,11 @@ def download_zip_file(url: str, output_dir: str) -> list:
     Returns:
       A list of files downloaded and unzipped.
     """
-    print(f'Downloading {url} to {output_dir}/', flush=True)
+    logging.info(f'Downloading {url} to {output_dir}/')
```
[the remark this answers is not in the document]
dhhhanushree: used download_util_script() with minor changes done in download_util_script()

scripts/un/energy/process.py
dhhhanushree: Uopdated

krishnaswamypradeep (approved):

scripts/un/energy/download.py
```diff
@@ -123,15 +102,28 @@ def download_energy_dataset(
...
+                                            delay=5,
+                                            backoff=2)
+        if download_successful:
+            logging.info(f"Download of '{download_url}' completed.")
+            output_files = []
+            for f in os.listdir(output):
+                output_files.append(os.path.join(output, f))
+                if len(output_files) == 0:
+                    errors += 1
+                output_files.extend(output_files)
```
n-h-diaz, line 124: Why both extend and append? Will this add the files multiple times?

scripts/un/energy/download.py
```diff
@@ -123,15 +102,28 @@ def download_energy_dataset(
...
+                                            headers=None,
+                                            tries=3,
+                                            delay=5,
+                                            backoff=2)
+        if download_successful:
+            logging.info(f"Download of '{download_url}' completed.")
+            output_files = []
+            for f in os.listdir(output):
+                output_files.append(os.path.join(output, f))
+                if len(output_files) == 0:
```
n-h-diaz, line 122: optional nit: can do something like 
if not output_files

Also, what is errors counting? It looks like output_files is getting re initialized every instance of the loop - should there instead be a separate variable for the per-url files?

scripts/un/energy/download.py
```diff
@@ -29,17 +29,22 @@
...
+sys.path.insert(1, os.path.join(_CODEDIR, '../../../util/'))
+from download_util_script import download_file
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string('download_data_dir', 'tmp_data_dir/un_energy',
+flags.DEFINE_string('download_data_dir', 'input_data/un_energy',
                     'Data dir to download into')
 flags.DEFINE_list('datasets', [], 'Datasets to download. Everything, if empty.')
-flags.DEFINE_integer('start_year', 1990,
+flags.DEFINE_integer('start_year', 2020,
```
n-h-diaz, line 47: Just curious, why did the start date change? Are we removing some of the data? Or is this a change from the source?

scripts/un/energy/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+    "import_specifications": [
+        {
+            "import_name": "UNEnergy",
+            "curator_emails": ["support@datacommons.org"],
+            "provenance_url": "https://unstats.un.org/unsd/energystats/data/",
+            "provenance_description": "UN_energy",
```
n-h-diaz, line 7: nit: could you make the description more readable? Like maybe a 1-2 sentence description of what this dataset contains

scripts/un/energy/process.py
```diff
@@ -533,8 +538,8 @@ def process(in_paths: list,
                         data_row['_Row'] = line
                         _process_row(data_row, sv_map, row_map, sv_obs,
                                      csv_writer, f_out_mcf, counters)
-                        _print_counters(counters, counters['debug_lines'])
-                print(f'Processed {line} rows from data file: {in_file}')
+                        # _print_counters(counters, counters['debug_lines'])
```
n-h-diaz, line 541: nit: remove instead of comment out - or is there a reason to keep this?

scripts/un/energy/process.py
```diff
@@ -479,6 +474,15 @@ def _process_row(data_row: dict, sv_map: dict, row_map: dict, sv_obs: dict,
     counters['output_csv_rows'] += 1
 
 
+def rename_the_mapping_value(sv_id, mapping_names):
```
n-h-diaz, line 477: nit: it would be helpful to have some function comment explaining what this does, as well as params, return value

scripts/un/energy/download.py
dhhhanushree: removed extend line

scripts/un/energy/manifest.json
dhhhanushree: Updated

scripts/un/energy/process.py
dhhhanushree: done

scripts/un/energy/download.py
n-h-diaz: Could you clarify or add a comment for what the errors variable is counting? Is it whether output is empty? (If so it won't get checked in the loop) Alternatively if the first combination of downloads is not empty, output_files won't be, so later combinations won't have the same check

n-h-diaz (approved):
Thanks for all the updates! 

I had one more quick clarification question, but overall looks good

PR #1260, opened 2025-03-04
Update executor python version

vish-cs:


import-automation/executor/cloudbuild.yaml
```diff
@@ -15,17 +15,25 @@ steps:
...
 
   # Run import executor integration test
-  - name: python:3.11.11
+  - name: python:3.12.8
     entrypoint: python
     args: ['verify_import.py']
     dir: 'import-automation/executor'
 
+  # Run unit tests
+  - name: 'python:3.12.8'
```
ajaits, line 30: I don't follow this with entrypoint being bash and name and args set to python. Shouldn't name be run_tests.sh and args, '-a'?

ajaits (approved):
Nice! no test updates required!

import-automation/executor/cloudbuild.yaml
vish-cs: name is the docker image we use so it should have python installed. Bash is the entry point to execute run_tests.sh shell script.

PR #1264, opened 2025-03-07
Data annotator and pvmap generator

ajaits:
Utilities for generating pv-map
- data_sampler.py: utility to down sample large csv files
- data_annotator.py: Annotate columns as place, date, or schema
- llm_pvmap_generator.py: Generate PV map from data strings.
- llm_name_generator.py: Generate statvar names


keyurva (approved):
I skimmed the PR and am approving it but would be good to do smaller PRs moving forward. One overall comment is that each script here is a runnable script. If it makes sense, consider exposing only one runner script and invoke others as modules as needed. This can simplify the individual modules and make them more testable.

ajaits:
> I skimmed the PR and am approving it but would be good to do smaller PRs moving forward. One overall comment is that each script here is a runnable script. If it makes sense, consider exposing only one runner script and invoke others as modules as needed. This can simplify the individual modules and make them more testable.

Thanks Keyur. 

While each script is can be run separately, all these modules are invoked by the stat_var_processor.py and don't need to be invoked individually.

PR #1281, opened 2025-03-27
Add GCS volume to executor container

vish-cs:
This PR adds a GCS bucket as a volume mount to the import executor docker container. Since the local storage for the import executor docker container is limited (cloud run uses memory as the local storage), GCS volume allows import scripts to download larger files to the external volume. The GCS bucket includes an import specific folder for each import to allow separation with a symlink to 'gcs_output' in the import script directory under data repo in the container.

### DIFF
--- a/import-automation/executor/app/executor/cloud_run.py
+++ b/import-automation/executor/app/executor/cloud_run.py
@@ -60,12 +60,21 @@ def create_or_update_cloud_run_job(project_id: str, location: str, job_id: str,
         env.append(run_v2.EnvVar(name=var, value=value))
 
     res = run_v2.types.ResourceRequirements(limits=resources)
-    container = run_v2.Container(image=image, env=env, resources=res, args=args)
+    mount = run_v2.types.VolumeMount(name='datcom-import-volume',
+                                     mount_path='/mnt')
+    source = run_v2.GCSVolumeSource(bucket='dc-import-volume')
+    volume = run_v2.types.Volume(name='datcom-import-volume', gcs=source)
+    container = run_v2.Container(image=image,
+                                 env=env,
+                                 resources=res,
+                                 args=args,
+                                 volume_mounts=[mount])
     # Labels allow filtering of automated import cloud run jobs, used in log-based metrics.
     exe_template = run_v2.ExecutionTemplate(
         labels={"datacommons_cloud_run_job_type": "auto_import_job"},
         template=run_v2.TaskTemplate(
             containers=[container],
+            volumes=[volume],
             max_retries=2,
             timeout=duration_pb2.Duration(seconds=timeout)))
     new_job = run_v2.Job(template=exe_template)
--- a/import-automation/executor/app/executor/cloud_run.py
+++ b/import-automation/executor/app/executor/cloud_run.py
@@ -28,8 +28,9 @@ from google.protobuf import duration_pb2
 
 
 def create_or_update_cloud_run_job(project_id: str, location: str, job_id: str,
-                                   image: str, env_vars: dict, args: list,
-                                   resources: dict, timeout: int) -> run_v2.Job:
+                                   image: str, mount_bucket: str,
+                                   env_vars: dict, args: list, resources: dict,
+                                   timeout: int) -> run_v2.Job:
     """Creates a new cloud run job or updates an existing one.
 
   If the jobs exists, the container is updated with new image and environment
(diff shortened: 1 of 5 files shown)


import-automation/executor/app/executor/cloud_run.py
```diff
@@ -60,12 +60,21 @@ def create_or_update_cloud_run_job(project_id: str, location: str, job_id: str,
         env.append(run_v2.EnvVar(name=var, value=value))
 
     res = run_v2.types.ResourceRequirements(limits=resources)
-    container = run_v2.Container(image=image, env=env, resources=res, args=args)
+    mount = run_v2.types.VolumeMount(name='datcom-import-volume',
+                                     mount_path='/mnt')
+    source = run_v2.GCSVolumeSource(bucket='dc-import-volume')
```
ajaits, line 65: If bucket and mount point can be the same name, can we call the bucket datcom-miunt-volume  too?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -455,6 +456,17 @@ def _invoke_import_job(self, absolute_import_dir: str, import_spec: dict,
         script_paths = import_spec.get('scripts')
         for path in script_paths:
             script_path = os.path.join(absolute_import_dir, path)
+            import_name = import_spec['import_name']
+            mount_path = os.path.join(self.config.gcs_volume_mount_dir,
+                                      import_name)
+            out_path = os.path.join(absolute_import_dir, 'output')
+            logging.info(f'Mount path: {mount_path}, Out path: {out_path}')
+            if os.path.exists(mount_path):
+                shutil.rmtree(mount_path)
```
ajaits, line 465: For scripts that need persistent files, can they create a custom folder under /mnt/... , say /mnt/persistent/import_name/... within the script?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -455,6 +456,17 @@ def _invoke_import_job(self, absolute_import_dir: str, import_spec: dict,
...
+            if os.path.lexists(out_path):
+                os.unlink(out_path)
+            os.makedirs(mount_path, exist_ok=True)
+            os.symlink(mount_path, out_path, True)
```
ajaits, line 469: Is this creating symlink from <import script  folder>/output  to the gcs volume mount ?  Some import scripts use current dir/output already but some don't. Maybe keep this output explicit , like, gcs_output? O/w it may lead to too many remote file io for small scripts that don't need volumes.

import-automation/executor/app/executor/cloud_run.py
vish-cs: Renamed both bucket and volume to datcom-mount-volume

import-automation/executor/app/executor/import_executor.py
vish-cs: Yes, scripts can use /mnt as they need

import-automation/executor/app/executor/import_executor.py
vish-cs: Renamed the symlink to gcs_output

import-automation/executor/app/executor/cloud_run.py
```diff
@@ -60,12 +60,21 @@ def create_or_update_cloud_run_job(project_id: str, location: str, job_id: str,
         env.append(run_v2.EnvVar(name=var, value=value))
 
     res = run_v2.types.ResourceRequirements(limits=resources)
-    container = run_v2.Container(image=image, env=env, resources=res, args=args)
+    mount = run_v2.types.VolumeMount(name='datcom-mount-volume',
+                                     mount_path='/mnt')
+    source = run_v2.GCSVolumeSource(bucket='datcom-mount-volume')
```
ajaits, line 65: can the GCS bucket name be configurable? custom DC may need a different bucket.

import-automation/executor/app/executor/import_executor.py
rohitkumarbhagat: Can we symlink /mnt/persistent/import_name/ folder within absolute import folder (similar to what we do for non-persistent one)? This might simplify usage by hiding the complexity of underlying infra from script writers.

import-automation/executor/app/executor/cloud_run.py
vish-cs: Updated to fetch bucket name from config.

ajaits (approved):
Thanks Vishal!

PR #1282, opened 2025-03-28
Add script to test imports

vish-cs:


import-automation/executor/import-test.sh
ajaits, line 1: Can we call this cloud_run_import.sh ?

ajaits (approved):

import-automation/executor/import-test.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Usage:
+# ./import-test.sh <job-name> <import-name>
+#
+# Example:
+# ./import-test.sh import-test scripts/us_fed/treasury_constant_maturity_rates:USFed_ConstantMaturityRates
+#
+
+# Customize these as per requirement
+config='{"gcs_project_id":"datcom-ci","storage_prod_bucket_name":"datcom-ci-test"}'
```
ajaits, line 11: Can we disable email notifications too?

vish-cs: Done

import-automation/executor/import-test.sh
vish-cs: Renamed to cloud_run_import_test.sh since this is only for testing purpose

PR #1284, opened 2025-03-28
Remove duplicates while adding a PV with list values to node

rohitkumarbhagat:


tools/statvar_importer/mcf_file_util_test.py
```diff
@@ -171,3 +171,120 @@ def test_get_numeric_value(self):
...
+    def test_add_pv_to_node_strip_namespaces(self):
+        node = {}
+        mcf_file_util.add_pv_to_node("prop1",
+                                     "dcid:value1",
+                                     node,
+                                     strip_namespaces=True,
+                                     normalize=False)
+        self.assertEqual(node, {"prop1": "value1"})
+
+    def test_add_pv_to_node_strip_namespaces(self):
```
ajaits, line 206: same as above?

tools/statvar_importer/mcf_file_util.py
```diff
@@ -613,12 +622,18 @@ def get_value_list(value: str) -> list:
...
-                       quotechar='"',
-                       skipinitialspace=True))
-        for v in rows[0]:
+        if '"' in value:
+            # Read the string as a quoted comma separated line.
+            row = list(
+                csv.reader([value],
+                           delimiter=',',
+                           quotechar='"',
+                           skipinitialspace=True))[0]
```
ajaits, line 631: perhaps also remember that value had quotes:
`is_quoted = True`

tools/statvar_importer/mcf_file_util.py
```diff
@@ -613,12 +622,18 @@ def get_value_list(value: str) -> list:
...
+        else:
+            # Without " quotes, the line can be split on commas.
+            # Avoiding csv reader calls for performance.
+            row = value.split(',')
+        for v in row:
             val_normalized = get_quoted_value(v)
```
ajaits, line 637: restore quotes if original value had quotes:
`get_quoted_value(v, is_quoted=is_quoted)`

tools/statvar_importer/mcf_file_util.py
```diff
@@ -613,12 +622,18 @@ def get_value_list(value: str) -> list:
     value_list = []
     # Read the string as a comma separated line.
     try:
-        rows = list(
-            csv.reader([value],
-                       delimiter=',',
-                       quotechar='"',
-                       skipinitialspace=True))
-        for v in rows[0]:
+        if '"' in value:
```
ajaits, line 625: can be a bit more faster by checking for both '"' and ','. csv reader is not needed if there are no ','.

tools/statvar_importer/mcf_file_util_test.py
rohitkumarbhagat: Updated to run with normalized=true

tools/statvar_importer/mcf_file_util.py
rohitkumarbhagat: Done

tools/statvar_importer/mcf_file_util.py
rohitkumarbhagat: done

tools/statvar_importer/mcf_file_util.py
rohitkumarbhagat: Good idea. Added the check.

PR #1294, opened 2025-04-03
updated the flag name of config

kurus21:
Corrected AttributeError: config in stat_var_processor.py (line 2995). The code was attempting to access a flag named config from _FLAGS, but the declared flag in config_flags.py is config_file. This commit updates the access to FLAGS.config_file to resolve the error.


krishnaswamypradeep (approved):

ajaits (approved):
Thanks for the quick fix!

PR #1295, opened 2025-04-03
code_fixes_mint

Avantika-Singh16:
Molecular Interaction Database (MINT) autorefresh configuration along with code fixes


scripts/biomedical/proteinInteractionMINT/download.py
```diff
@@ -0,0 +1,34 @@
+import requests
```
krishnaswamypradeep, line 1: Please use retry module with retry options.

PR #1307, opened 2025-04-08
Statvar_download_script

SwaroopN-123:


util/statvar_download_util.py
```diff
@@ -0,0 +1,89 @@
...
+from statvar_download_util import StatVarDownloader
+
+downloader = StatVarDownloader()
+
+urls = [
+    "https://example.com/data.csv",
+    "https://example.com/data/report.csv"
+]
+
+downloader.download_files(urls, "/your/output/folder")
```
krishnaswamypradeep, line 17: What is the intended return type of this method? Also, how will consumers of this utility be informed about the success or failure of the file download?

util/statvar_download_util.py
```diff
@@ -0,0 +1,89 @@
...
+
+import os
+import time
+import hashlib
+import zipfile
+from io import BytesIO
+from typing import List
+from urllib.parse import urlparse
+import requests
+import pandas as pd
```
krishnaswamypradeep, line 32: why pandas user here?

PR #1312, opened 2025-04-10
Add volume support for local docker builds

vish-cs:


import-automation/executor/cloud_run_import_test.sh
```diff
@@ -11,18 +11,23 @@
...
 #
 # Customize these as per requirement
-config='{"gcs_project_id":"datcom-ci","storage_prod_bucket_name":"datcom-ci-test","disable_email_notifications":true}'
-project=datcom-ci
+config='{"gcs_project_id":"datcom-infosys-dev","storage_prod_bucket_name":"datcom-import-test","disable_email_notifications":true}'
+project=datcom-infosys-dev
 region=us-west1
+artifact_registry=us-central1-docker.pkg.dev/datcom-infosys-dev/datcom-infosys-dev-artifacts
+mount_bucket=datcom-import-test
+data_repo=$HOME/data
```
ajaits, line 19: can we derive this from the script path?
```
script_path=$(dirname $0)
data_repo=$(realpath $script_path/../../)
```

ajaits (approved):

import-automation/executor/cloud_run_import_test.sh
```diff
@@ -11,18 +11,23 @@
 # ./cloud_run_import_test.sh import-test scripts/us_fed/treasury_constant_maturity_rates:USFed_ConstantMaturityRates
 #
 # Customize these as per requirement
-config='{"gcs_project_id":"datcom-ci","storage_prod_bucket_name":"datcom-ci-test","disable_email_notifications":true}'
-project=datcom-ci
+config='{"gcs_project_id":"datcom-infosys-dev","storage_prod_bucket_name":"datcom-import-test","disable_email_notifications":true}'
```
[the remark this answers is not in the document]
vish-cs: datcom-ci-test is a test bucket used for integration tests. It is not a prod bucket.

import-automation/executor/cloud_run_import_test.sh
ajaits: If the $HOME/data doesn't exist does it fallback to the git repo clone?

Since this script is for testing local changes, deriving the local dir having a way to pick local changes would be useful without having to edit it. 

o/w this script will have to be always edited before use since the local folder will likely be different for each user.

vish-cs: I have changed the default to use the local data repo.

PR #1314, opened 2025-04-10
WB_CommodityMarket StatVar configuration update

kurus21:
Files added:
PV map
metadata
download script to download input files/data files
manifest json
run.sh file that runs download and stat_var processor script and the generates output file.


kurus21:
Divergence issue.

PR #1323, opened 2025-04-16
Fix CodeQL workflow by granting write permissions on security-events

clincoln8:


hqpho:
We did this for website too in https://github.com/datacommonsorg/website/pull/4889 . Do we need to do any work for other repos?

hqpho (approved):

clincoln8:
Thanks for the speedy review!!

PR #1334, opened 2025-04-22
USCensusPEP_Sex cloud automation update

kurus21:
The manifest json has been updated.
timeout error has been handled in the process.py


scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -881,8 +881,8 @@ def _county_latest(file_path: str) -> pd.DataFrame:
                 4: '2022',
                 5: '2023',
                 6: '2024',
-                7: '2025',
-                8: '2026',
+                7: '2030',
```
krishnaswamypradeep, line 884: Why year 2025 & 2026 removed?

kurus21: Removed by mistake while using find and replace.
Corrected it.

PR #1343, opened 2025-04-29
Support wildcard in manifest.json cleaned_csv

ajaits:
Lookup all files for cleaned_csv with wildcards and pass it to dc-import lint for validation.


krishnaswamypradeep (approved):
Hi Ajai, We have  tested this and it is working as expected.

PR #1349, opened 2025-04-30
Scripts to test imports locally or on cloud run

ajaits:
Adds a script to run test imports:
- locally within a docker
- or on cloud run

Also updates existing scripts to set configs for test imports to disable updating prod GCS buckets.


import-automation/executor/config_override_test.json
```diff
@@ -0,0 +1,12 @@
+{
+    "configs":{
+        "executor_type": "CLOUD_RUN",
+        "user_script_args": [],
+        "user_script_env": {},
+        "user_script_timeout": 3600,
+        "disable_email_notifications": true,
+        "import_tool_path" : "/tmp/import-tool/import-tool.jar",
```
vish-cs, line 8: Do we also need to add differ jar since differ is also added now? We can also have a common variable to store the path and hard code the binary names for both.

import-automation/executor/config_override_test.json
```diff
@@ -0,0 +1,12 @@
...
+        "gcs_volume_mount_dir": "/tmp",
+        "storage_prod_bucket_name": "datcom-import-test"
```
vish-cs, line 10: do we also need project id?

import-automation/executor/run_import.sh
```diff
@@ -0,0 +1,298 @@
+#!/bin/bash
+#
+# Prerequisites:
+# - Install gcloud, docker
+# - Run 'glcoud auth login'
```
vish-cs, line 5: I wonder if we also need to run 'gcloud auth application-default login'? docker run is using ADC.

import-automation/executor/run_import.sh
```diff
@@ -0,0 +1,298 @@
...
+DEFAULT_MEMORY=4Gi
+DEFAULT_TIMEOUT=30m
+RUN_MODE="local"
+DOCKER_IMAGE="dc-import-executor"
+CONFIG="config_override_test.json"
+TMP_DIR=${TMP_DIR:-"/tmp"}
+USAGE="$(basename $0) <import-name> [Options]
+Script to run an import through docker in cloud run or locally.
+Options:
+  -local       Run the import locally.
```
vish-cs, line 28: -local and -docker can be confusing since they both mention about running locally. One option is to give them different names (native vs docker).

import-automation/executor/run_import.sh
```diff
@@ -0,0 +1,298 @@
...
+  -i <import-name> Run the import for <import-name> in the manifest.json.
+  -p <project> GCP project. Default: $GCP_PROJECT
+  -region <region>  Region. Default: $REGION
+  -a <reg>     Arifact registry: Default:gcr.io/$GCP_PROJECT
+  -b <bucket>  GCS bucket to be mounted in the job. Default:$GCS_BUCKET
+  -repo <dir>  Data repo directory. Default:$DATA_REPO
+  -cpu <N>       Number of CPUs for the job. Default:$DEFAULT_CPU
+  -mem <M>       Memory. Default:$DEFAULT_MEMORY
+  -timeout <secs>    Timeout. Default:$DEFAULT_TIMEOUT
+  -config <json> JSON file wiht import executor configs. Default: $CONFIG
```
vish-cs, line 42: typo: with

import-automation/executor/run_import.sh
```diff
@@ -0,0 +1,298 @@
...
+DOCKER_IMAGE="dc-import-executor"
+CONFIG="config_override_test.json"
+TMP_DIR=${TMP_DIR:-"/tmp"}
+USAGE="$(basename $0) <import-name> [Options]
+Script to run an import through docker in cloud run or locally.
+Options:
+  -local       Run the import locally.
+  -cloud       Run import as a cloud run job.
+  -docker      Run import locally through docker.
+  -d <image-name>  Build a docker image wiht tage and run the docker.
```
vish-cs, line 31: typo: with tag

import-automation/executor/run_import.sh
vish-cs: I checked the rest of the PR. Seems we are setting up ADC later so the comment can be ignored.

import-automation/executor/run_import.sh
```diff
@@ -0,0 +1,298 @@
...
+  MEMORY=$(get_param_value_json "memory" "$MEMORY" "$MANIFEST" "$DEFAULT_MEMORY")
+  TIMEOUT=$(get_param_value_json "user_script_timeout" "$TIMEOUT" "$MANIFEST" "$DEFAULT_TIMEOUT")
+
+  echo_log "Loaded manifest for $IMPORT_DIR:$IMPORT_NAME"
+}
+
+# Build a docker image locally if needed
+function build_docker {
+  [[ ! -f $HOME/config/gcloud/application_default_credentials.json ]] || \
+    run_cmd gcloud auth application-default login
```
vish-cs, line 170: This step requires user interaction through UI so it might have to be done during pre-requisites?

import-automation/executor/config_override_test.json
ajaits: Added the differ jar too. Let's keep the seperate variables fr now. We may change this to a released version once we build with cloud build.

import-automation/executor/config_override_test.json
ajaits: Added gcp_project_id

import-automation/executor/run_import.sh
ajaits: changed -local to -executor

import-automation/executor/run_import.sh
ajaits: Done.

import-automation/executor/run_import.sh
ajaits: Done

vish-cs (approved):

PR #1351, opened 2025-05-02
US PEP Sex Cloud Fix 20250429

kurus21:


scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1156,52 +1184,102 @@ def add_future_year_urls():
...
+            with session.get(url, stream=True, timeout=120,
+                             headers=headers) as response:
+                response.raise_for_status()
+
+                content_type = response.headers.get('Content-Type', '')
+                with open(_INPUT_URL_JSON, 'r') as f:
+                    json_data = json.load(f)
+                if url in json_data:
+                    if 'html' in content_type.lower():
+                        logging.fatal(
```
krishnaswamypradeep, line 1237: After a `logging.fatal(`) call, the script will terminate; the `continue` statement will not be executed. Could you please review the loop's logic again?

kurus21: Yes right. Thanks for it.
Removed the continue. The loop supposed to break if it was catched in that 'if' condition.

krishnaswamypradeep (approved):
Thanks Kuru

PR #1353, opened 2025-05-08
USCensusPEP_Sex Persistent Folder addition

kurus21:
The process script was enhanced with the following functionality:

Files were downloaded to a temporary subfolder inside the existing gcs_output folder (created externally by Docker).

In case of a partial download failure, the script:

Detected and listed the already downloaded files in the temporary folder.

Skipped these already downloaded files on the subsequent run.

Resumed downloading the remaining files from where the previous attempt had failed.

This approach aimed to:

Improve the reliability and success rate of imports.

Prevent exhaustion of the maximum retry limit imposed by the source website.


scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1206,10 +1211,18 @@ def download_files():
 
     skipped = []
 
+    # Step 1: Get set of already downloaded files
+    downloaded_files = set(os.listdir(_GCS_OUTPUT))
+
     for file_to_download in _FILES_TO_DOWNLOAD:
```
krishnaswamypradeep, line 1217: Please refactor the code for generating the `_FILES_TO_DOWNLOAD` variable as follows:

 1. **Prepare all download URLs.**
 2. **Filter/skip URLs based on the configuration file.**

This approach will eliminate the need for redundant validation during the actual download process.

scripts/us_census/pep/us_pep_sex/manifest.json
```diff
@@ -6,7 +6,7 @@
       "provenance_url": "https://www2.census.gov/programs-surveys/popest/tables",
       "provenance_description": "US Census Population Estimates Program (PEP).",
       "scripts": ["process.py --config_path=unresolved_mcf/us_census/pep/population_estimate_by_sex/skip_url.json"],
-      "source_files": ["input_files/*"],
+      "source_files": ["gcs_output/us_pep_sex_source_files/*"],
```
krishnaswamypradeep, line 9: To enable persistent volumes in Cloud run, add the property 
`"cleanup_gcs_volume_mount": false`

scripts/us_census/pep/us_pep_sex/process.py
kurus21: Restructed

scripts/us_census/pep/us_pep_sex/manifest.json
kurus21: Added.

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1322,6 +1345,12 @@ def main(_):
                                          mcf_path, tmcf_path)
         loader.process()
 
+        # Only delete if it's a subdirectory of gcs_output, and not gcs_output itself
```
krishnaswamypradeep, line 1348: Add a try...except block around the loader.process() call. 
This is crucial because loader.process() can return None without raising an explicit exception, and there's currently no handling for this potential scenario. 

This could lead to unexpected behavior later in the execution

```
try:
        loader = PopulationEstimateBySex(_INPUT_FILE_PATH, cleaned_csv_path,
                                         mcf_path, tmcf_path)
        loader.process()

        # Only delete if it's a subdirectory of gcs_output, and not gcs_output itself
        if os.path.exists(_GCS_OUTPUT_PERSISTENT_PATH) and os.path.commonpath([_GCS_OUTPUT_PERSISTENT_PATH, _GCS_BASE_DIR]) == _GCS_BASE_DIR \
        and os.path.abspath(_GCS_OUTPUT_PERSISTENT_PATH) != os.path.abspath(_GCS_BASE_DIR):
            shutil.rmtree(_GCS_OUTPUT_PERSISTENT_PATH)
            logging.info(f"Deleted folder: {_GCS_OUTPUT_PERSISTENT_PATH}")

catch Exception as e:
        logging.fatal(f"Error occurred {e}")

```

kurus21: Added. Thanks

PR #1354, opened 2025-05-08
FBIgovcrime-Fix given for selenium and docker issue

dhhhanushree:
Fix given for selenium and docker issue

### DIFF
--- a/import-automation/executor/Dockerfile
+++ b/import-automation/executor/Dockerfile
@@ -19,7 +19,23 @@ ARG build_type=cloud
 
 # build base image with all the dependencies installed
 FROM python:3.12.8 as base
-RUN apt-get update && apt-get install -y jq
+RUN apt-get update && apt-get install -y jq \
+curl \
+wget \
+gnupg \
+apt-transport-https \
+ca-certificates \
+libglib2.0-0 \
+libnss3 \
+libgconf-2-4 \
+libatk1.0-0 \
+libgtk-3-0 \
+libx11-6 \
+libasound2 \
+fonts-liberation \
+xdg-utils \
+chromium \
+chromium-driver
 ENV JAVA_HOME=/usr/local/openjdk-17
 COPY --from=openjdk:17-slim $JAVA_HOME $JAVA_HOME
 ENV PATH="${JAVA_HOME}/bin:${PATH}"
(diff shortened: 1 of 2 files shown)


import-automation/executor/Dockerfile
```diff
@@ -19,7 +19,23 @@ ARG build_type=cloud
 
 # build base image with all the dependencies installed
 FROM python:3.12.8 as base
-RUN apt-get update && apt-get install -y jq
+RUN apt-get update && apt-get install -y jq \
+curl \
```
vish-cs, line 23: Many packages like curl wget are already in the base docker image so we don't have to explicitly install them. Can we check which ones are really missing?

import-automation/executor/Dockerfile
```diff
@@ -19,7 +19,23 @@ ARG build_type=cloud
...
+libnss3 \
+libgconf-2-4 \
+libatk1.0-0 \
+libgtk-3-0 \
+libx11-6 \
+libasound2 \
+fonts-liberation \
+xdg-utils \
+chromium \
+chromium-driver
```
vish-cs, line 38: What is the size of the resultant docker image after adding these packages?

import-automation/executor/Dockerfile
vish-cs: Build the image locally
$ DOCKER_BUILDKIT=1 docker build -t dc-executor .

Run it with bash as entry point
$ docker run -it --entrypoint "/bin/bash"  dc-executor

It will launch the docker container and drop you into a bash shell. Then you can try any command at the prompt (curl) to see if it's already included.

import-automation/executor/Dockerfile
vish-cs: Once you build the image locally, you can run "docker image ls" to check the size.

prajapata: After adding these packages the size of the docker image is 5.09GB. Please find the attached screenshot: https://screenshot.googleplex.com/9goKMuZTNsC2g8J

import-automation/executor/Dockerfile
prajapata: curl and wget were removed, and testing was successful.

vish-cs (approved):

PR #1384, opened 2025-05-27
Add native python differ mode

vish-cs:


tools/import_differ/differ_utils.py
```diff
@@ -144,3 +144,17 @@ def get_gcs_data(uri: str, tmp_dir: str) -> str:
...
+
+def load_data(path: str, tmp_dir: str) -> pd.DataFrame:
+    """ Loads data from the given path and returns as a dataframe.
+    Args:
+      path: local or gcs path (single file or wildcard format)
+      tmp_dir: destination folder
+    Returns:
+      dataframe with the input data
+    """
+    os.makedirs(tmp_dir, exist_ok=True)
```
ajaits, line 157: os.makedirs() can be within the if condition below?

tools/import_differ/import_differ.py
```diff
@@ -125,11 +125,66 @@ def _get_samples(self, row):
...
+          current_df: dataframe with current (new) data
+          previous_df: dataframe with previous (old) data
+        Returns:
+          intermediate merged data for analysis
+        """
+        cur_df_columns = current_df.columns.values.tolist()
+        for column in self.groupby_columns:
+            if column not in cur_df_columns:
+                current_df[column] = ''
+                previous_df[column] = ''
```
ajaits, line 144: reset column to empty only if column doesn't exist? that way if a new column is added in the future, diff can still run without ignoring it?

tools/import_differ/differ_utils.py
vish-cs: Done

tools/import_differ/import_differ.py
vish-cs: Isn't the current logic already doing it...we reset the column only if is 'not' in cur_df_columns?

PR #1395, opened 2025-06-09
African data download issue in cloud run fix

krishnaswamypradeep:
This PR contains a fix to address the XML parser errors encountered during African dataset downloads in the Cloud Run environment. We'll monitor its performance and, if the solution doesn't work as expected, we'll initiate a rollback


manishvats2 (approved):
approved

PR #1419, opened 2025-06-17
download and process script

Avantika-Singh16:


scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
```diff
@@ -0,0 +1,205 @@
...
+    "output_csv_name":
+        "ncses_employed_female.csv"
+}]
+
+# Construct the full path to the download_util_script.py
+download_script_path = os.path.join(resolved_util_dir,
+                                    'download_util_script.py')
+
+# Ensure the download script exists before proceeding
+if not os.path.exists(download_script_path):
```
krishnaswamypradeep, line 32: why is this check required here?

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
krishnaswamypradeep, line 1: 1. Copyright block is missing
2. Please use absl app to invoke the main method
3. Avoid print statements 
4. Update the PR name as per the format we follow
5. Use the standard way to import util path

```
# Add PATH to import modules from data/util
_SCRIPT_PATH = os.path.dirname(os.path.abspath(__file__))
sys.path.append(_SCRIPT_PATH)
sys.path.append(_SCRIPT_PATH.split('/data/', 1), 'data', 'util')

```

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
```diff
@@ -0,0 +1,205 @@
...
+
+        logging.info(f"\n--- Attempting to download: {url} ---")
+
+        # Extract filename from URL to know what to expect after download
+        file_name_from_url = url.split('/')[-1]
+        downloaded_file_path = os.path.join("temp_downloads",
+                                            file_name_from_url)
+
+        # Define the command for the current URL
+        command = [
```
krishnaswamypradeep, line 137: Why subprocess is required here? You can directly call the  download until method here

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
krishnaswamypradeep, line 1: Can you raise PR with PV map & configs files all at once?

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
Avantika-Singh16: This check is not compulsory so removed this if block

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
Avantika-Singh16: Resolved

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
Avantika-Singh16: yes we can directly call the function.I have removed the sub process

scripts/NCSES_Research_Doctorate_Recipients_Import/download_and_process.py
Avantika-Singh16: Added

PR #1441, opened 2025-06-30
NCES_Bachelors_Degree_By_Field_Import Configs update

SudhishaK:


statvar_imports/us_bachelors_degree_data/nces_bachelors_degree_by_field_import/download_data.py
```diff
@@ -41,13 +41,13 @@ def download(inputdir):
                     logging.warning(f"Skipping download for {url}. Received non-200 status code: {response.status_code}")
                 count+=1
             except Exception as e:
-                logging.error(f"Error occured while downloading the data: {e}")
```
hareesh-ms, line 44: The year replacement in url is done twice in line 32,34 which seems redundant.  Update if required.

hareesh-ms (approved):

PR #1462, opened 2025-07-11
Fix mount point to match cloud run config

vish-cs:


import-automation/executor/app/executor/import_executor.py
```diff
@@ -526,9 +526,7 @@ def _invoke_import_job(self, absolute_import_dir: str, import_spec: dict,
                            process: subprocess.CompletedProcess) -> None:
         script_paths = import_spec.get('scripts')
         import_name = import_spec['import_name']
-        if not (mount_path := self.config.gcs_volume_mount_dir):
-            mount_path = os.path.join(self.config.local_repo_dir, 'mnt')
-        self._create_mount_point(mount_path,
+        self._create_mount_point(self.config.gcs_volume_mount_dir,
```
ajaits, line 529: When running locally should the gcs_volume_mount_dir always be set now as the default '/mnt' would not be accessible locally.
Shoudl we modify run_import.sh to override this when running locally?

PR #1464, opened 2025-07-11
Ignore differ output for first run of import

ajaits:
When running an import for the first time, there is no previous output to compare, so differ is skipped.
When validation the import, ignore the differ output when it is not generated.

### DIFF
--- a/tools/import_validation/import_validation.py
+++ b/tools/import_validation/import_validation.py
@@ -73,7 +73,7 @@ class ImportValidation:
             self.differ_results = pd.read_csv(differ_output)
             print('Differ results\n:', self.differ_results)
         else:
-            logging.info(f'No differ resuults loaded')
+            logging.info(f'No differ results loaded')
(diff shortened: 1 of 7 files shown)


tools/import_validation/import_validation.py
```diff
@@ -68,8 +68,14 @@ class ImportValidation:
...
                  stats_summary: str, validation_output: str):
         logging.info('Reading config from %s', validation_config)
-        self.differ_results = pd.read_csv(differ_output)
-        print(self.differ_results)
+        if differ_output:
+            logging.info(f'Loading differ results from {differ_output}')
+            self.differ_results = pd.read_csv(differ_output)
+            print('Differ results\n:', self.differ_results)
+        else:
+            logging.info(f'No differ resuults loaded')
```
vish-cs, line 76: typo: results

vish-cs (approved):

PR #1473, opened 2025-07-15
refactor: Add unit tests for PropertyValueCache class

SandeepTuniki:
(Part of a series of general improvements I am doing with the help of Gemini CLI. These changes do not introduce any new functionality but make the existing code easier to understand, use, and maintain.)

This PR extends a rudimentary set of unit tests for the `property_value_cache.py` module to enable safe refactoring and improve its reliability. I will extend it with more tests in a future PR, and use them to refactor the existing code.

I specifically focused on testing the methods that are relied on by the `PlaceResolver` and `WikiPlaceResolver` classes.


tools/statvar_importer/property_value_cache_test.py
```diff
@@ -22,106 +25,349 @@
...
+        """Tests failed entry lookup for a non-existent value."""
+        pv_cache = PropertyValueCache()
+        entry = {'dcid': 'geoId/06', 'name': 'California'}
+        pv_cache.add(entry)
+
+        # Lookup by a value that does not exist
+        self.assertEqual({}, pv_cache.get_entry('non-existent-value'))
+
+    @unittest.skip(
+        "TODO: The cache currently merges entries with conflicting key property values instead of overwriting them. The desired behavior needs to be determined."
```
SandeepTuniki, line 120: @ajaits Could you review this test specifically please? I am not sure if the behaviour of the `add` method should be to override or extend the existing entries.

ajaits: What happens if there are add() calls with different dcids?

```
pv_cache.add({'name': 'USA', 'dcid': 'country/USA'})
pv_cache.add({'name': 'USA', 'dcid': 'wikidataId/Q30'})
```

pv_cache.get_entry('country/USA')

PR #1530, opened 2025-08-07
executor

ajaits:
Abort import if the user script output files are not found.

Also skip update to  latest_version.txt if gcs_copy is disabled.


import-automation/executor/config_override.json
```diff
@@ -1,7 +1,7 @@
 {
     "configs":{
-        "executor_type": "CLOUD_BATCH",
+        "executor_type": "CLOUD_RUN",
```
vish-cs, line 3: This was accidentally left out from the last PR. thanks for fixing.

vish-cs (approved):

PR #1577, opened 2025-08-25
fix(validation): Handle empty and missing differ files

SandeepTuniki:
This PR fixes a bug (internal: b/440969217) in the import validation framework where it would crash or skip validations when the differ output file was missing or empty.

**Changes:**

- The `ValidationRunner` no longer raises an exception when a differ file is required but not provided. It now logs a warning and proceeds with an empty DataFrame, allowing validations to run without crashing.
- Validations are no longer skipped for empty DataFrames, ensuring that checks like `DELETED_COUNT` run correctly on empty or missing differ files.
- The `Validator`'s differ-related functions were updated to correctly handle DataFrames with no columns, which occurs when a differ file is missing.
- New tests were added to cover these specific scenarios.

### DIFF
--- a/import-automation/executor/app/configs.py
+++ b/import-automation/executor/app/configs.py
@@ -122,7 +122,7 @@ class ExecutorConfig:
     # Cloud workflow id.
     cloud_workflow_id: str = 'import-automation-workflow'
     # Maximum time a user script can run for in seconds.
-    user_script_timeout: float = 36000
+    user_script_timeout: float = 86400
     # Arguments for the user script
     user_script_args: List[str] = ()
     # Environment variables for the user script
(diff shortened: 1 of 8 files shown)


tools/import_validation/validator.py
```diff
@@ -141,14 +141,22 @@ def validate_deleted_count(self, differ_df: pd.DataFrame,
...
-        if 'DELETED' not in differ_df.columns:
-            return ValidationResult(
-                ValidationStatus.DATA_ERROR,
-                'DELETED_COUNT',
-                message="Input data is missing required column: 'DELETED'.")
-
-        rows_processed = len(differ_df)
         if differ_df.empty:
+            deleted_count = 0
+            threshold = params.get('threshold', 0)
```
vish-cs, line 146: Don't we also need the same threshold logic for other checks (ADDED/MOFIIED)?

tools/import_validation/runner.py
```diff
@@ -82,13 +82,9 @@ def _initialize_data_sources(self, stats_summary: str, differ_output: str):
             )
 
         if differ_required:
-            if differ_output and not os.path.exists(differ_output):
-                raise ValueError(
-                    f"A validation rule requires the 'differ' data source, but the --differ_output file was not provided or does not exist. Path: {differ_output}"
-                )
-            if not differ_output:
+            if not differ_output or not os.path.exists(differ_output):
```
vish-cs, line 85: If the file is not present...we log the warning. Do we still run the validation? what is the status that is returned?

PR #1580, opened 2025-08-25
Fix default PV retrieval in StatVarDataProcessor

rohitkumarbhagat:
Use default_pv_string instead of normalize_filename when getting default PVs.


ajaits (approved):
Thank your spotting this!

PR #1614, opened 2025-09-15
Add spanner ingestion workflow

vish-cs:
This PR adds a GCP workflow (spanner-ingestion-workflow) to perform incremental ingestion of spanner graph Data. It makes use of a cloud function (spanner-ingestion-helper) which handles state management in spanner tables. The core data ingestion to spanner graph is performed by a dataflow pipeline. 

The workflow performs the following high-level tasks:
 - Acquire ingestion lock
 - Get the list of pending imports
 - Run dataflow ingestion job
 - Update import status in spanner
 - Release ingestion lock

### DIFF
--- a/import-automation/workflow/helper/spanner_client.py
+++ b/import-automation/workflow/helper/spanner_client.py
@@ -1,0 +1,160 @@
+import logging
+from google.cloud import spanner
+from google.cloud.spanner_v1 import Transaction
+from google.cloud.spanner_v1.param_types import STRING, TIMESTAMP, Array
+from datetime import datetime, timezone
+
+logging.getLogger().setLevel(logging.INFO)
+
+
+class SpannerClient:
+    """
+    Spanner client to handle tasks like acquiring/releasing lock 
+    and getting/updating import statuses.
+    """
+    _LOCK_ID = "global_ingestion_lock"
+
+    def __init__(self, project_id: str, instance_id: str, database_id: str):
+        """Initializes a Spanner client and connects to a specific database."""
+        spanner_client = spanner.Client(
+            project=project_id, client_options={'quota_project_id': project_id})
+        instance = spanner_client.instance(instance_id)
+        database = instance.database(database_id)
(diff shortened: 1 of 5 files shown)


import-automation/workflow/spanner-ingestion-workflow.yaml
```diff
@@ -0,0 +1,156 @@
...
+# - Get the list of pending imports
+# - Run dataflow ingestion job
+# - Update import status in spanner
+# - Release ingestion lock
+spanner_ingestion_workflow:
+  params: [workflow_id]
+  steps:
+    - init:
+        assign:
+          - lock_timeout: 3600 #seconds
```
n-h-diaz, line 40: Just to double check, what will happen if the lock times out before a job completes? Is 1 hr sufficient for larger imports?

import-automation/workflow/helper/spanner_client.py
```diff
@@ -0,0 +1,158 @@
...
+                },
+                param_types={
+                    "importNames": Array(STRING),
+                    "workflowId": STRING
+                })
+            logging.info(f"Marked {updated_rows} import jobs as DONE.")
+
+            # 2. Insert into the IngestionHistory table
+            insert_sql = """
+                INSERT INTO IngestionHistory (CompletionTimestamp, WorkflowExecutionID, IngestedImports)
```
n-h-diaz, line 143: To confirm, are the recent rows with empty IngestedImports failed dataflow jobs?

vish-cs:
> This is really cool!
> 
> A few quick questions for my understanding:
> 
> * do you have a sense for how much spanner compute this will require?
> * how long does bulk deletion take?
> * will non autorefresh imports also be set up with a "pending" schedule for writing to spanner?

Thanks for the review!

We need to run this on large import to get a realistic measure of spanner load.
Bulk deletion logic is yet to be added to the dataflow pipeline. I will do that next.
Similarly, handling non auto refresh imports is TBD. There are multiple options to handle this. I will share a design doc on it.

PR #1624, opened 2025-09-18
Utility to extract relevant SDMX dataflow metadata

rohitkumarbhagat:
agentic import, extracted metadata could be read by Gemini in case xml metadata is too big for context.


tools/sdmx_import/sdmx_metadata_extractor.py
```diff
@@ -0,0 +1,450 @@
...
+class ReferencedConceptSchemeDetails:
+    """Represents a ConceptScheme and its Concepts referenced within the Dataflow."""
+    id: str
+    name: str = ""
+    description: str = ""
+    concepts: List[ConceptDetails] = field(default_factory=list)
+
+
+@dataclass
+class DataflowSomeAttributes:
```
ajaits, line 144: instead of `Some` can we use `Artefact`

ajaits (approved):

tools/sdmx_import/sdmx_metadata_extractor.py
rohitkumarbhagat: Done. Thanks for catching this,

PR #1644, opened 2025-09-30
Update import integration test to use cloud batch

vish-cs:


import-automation/executor/app/executor/cloud_batch.py
```diff
@@ -165,3 +169,91 @@ def create_job_request(import_name: str, import_config: dict, import_spec: dict,
...
+    allocation_policy = batch_v1.AllocationPolicy()
+    allocation_policy.instances = [instances]
+
+    job = batch_v1.Job()
+    job.task_groups = [group]
+    job.allocation_policy = allocation_policy
+    job.logs_policy = batch_v1.LogsPolicy()
+    job.logs_policy.destination = batch_v1.LogsPolicy.Destination.CLOUD_LOGGING
+
+    create_request = batch_v1.CreateJobRequest()
```
ajaits, line 239: should we move create_job_request() also to use batch_v1?

import-automation/executor/verify_import.py
```diff
@@ -13,31 +13,44 @@
...
+import time
 
-from app.executor import cloud_run
+from app.executor import cloud_batch
 from app.executor import file_io
 from test import utils
 
 _PROJECT_ID = 'datcom-ci'
-_JOB_ID = 'dc-import-prober'
+_JOB_ID_PREFIX = 'dc-import-itest'
```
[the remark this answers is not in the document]
vish-cs: It's supposed to be integration test but I can make it dc-import-test to avoid confusion

import-automation/executor/app/executor/cloud_batch.py
vish-cs: Actually create job request does not create a new batch job request but creates the json payload to trigger a cloud batch job from a cloud scheduler job/cloud workflow. Thus, we are not directly using batch_v1.

PR #1655, opened 2025-10-07
Add --gemini_cli flag for custom CLI command

rohitkumarbhagat:
Enables users to specify custom Gemini CLI invocation (aliases, custom paths) when gemini is not in PATH or requires special execution context.

### DIFF
--- a/import-automation/executor/run_import.sh
+++ b/import-automation/executor/run_import.sh
@@ -27,11 +27,11 @@
 # Example:
 # ./run_import.sh ../../scripts/us_fed/treasury_constant_maturity_rates/manifest.json
 #
-GCP_PROJECT="datcom-ci"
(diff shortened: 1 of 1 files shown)


tools/agentic_import/pvmap_generator.py
```diff
@@ -67,6 +67,11 @@
     'output_path', 'output/output',
     'Output path prefix for all generated files (default: output/output)')
 
+flags.DEFINE_string(
+    'gemini_cli', None, 'Custom path or command to invoke Gemini CLI. '
```
ajaits, line 71: can we default to 'gemini' instead of None?

ajaits (approved):

rohitkumarbhagat (commented):
addressed comments

PR #1893, opened 2026-02-16
fixed process.py in UsCensuspep_sex import 

niveditasing:
Code fix- The code involves a reverse-chronological crawl (starting from 2030) , to resolve this and optimize performance, implemented  with new "gatekeeper" logic: the script now validates the primary URL for a year and, if missing, skips that year entirely to reduce execution time. Additionally, I integrated a check to identify and skip HTML error pages returned by the server, preventing the  pipeline crash.

After  code fix the pipleine is working fine & giving expected output
image


saanikaaa (approved):

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1255,6 +1273,12 @@ def download_files():
         else:
             file_name_to_save = url.split('/')[-1]
 
+        # Skip if file already exists (Moved up for efficiency)
+        if file_name_to_save in downloaded_files:
```
ajaits, line 1277: Is there a TTL for file sin persistent GCS so that files that are too old, say older than a month, aren't reused but are downloaded again?

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1263,45 +1287,39 @@ def download_files():
...
+
+                if response.status_code == 200:
+                    with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+                        for chunk in response.iter_content(chunk_size=8192):
+                            if chunk:
+                                tmp_file.write(chunk)
+                        tmp_file_path = tmp_file.name
+
+                    # Copy to local destination
+                    shutil.copy(
```
ajaits, line 1305: why not shutil.move() which could be faster for local files within the same drive

scripts/us_census/pep/us_pep_sex/process.py
niveditasing: Added a _TTL_DAYS = 30 constant. The download_files function now checks the age of existing files in
     the persistent storage; if a file is older than 30 days, it is re-downloaded. This ensures we don't rely on stale data if the source is
     updated.

scripts/us_census/pep/us_pep_sex/process.py
niveditasing: Switched from shutil.copy() + os.remove() to shutil.move()

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -45,6 +45,7 @@
 _GCS_OUTPUT_PERSISTENT_PATH = os.path.join(
     _MODULE_DIR, 'gcs_output/us_pep_sex_source_files')
```
ajaits: Please use `gcs_folder/` instead of `gcs_output/`

gcs_output is now just a local folder not persisted in GCS.

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -45,6 +45,7 @@
...
 _GCS_BASE_DIR = os.path.join(_MODULE_DIR, 'gcs_output')
```
ajaits: please change this to `gcs_folder` or remove if unused.

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1255,6 +1277,20 @@ def download_files():
...
+        if file_name_to_save in downloaded_files:
+            file_path = os.path.join(_GCS_OUTPUT_PERSISTENT_PATH,
+                                     file_name_to_save)
+            file_age = (time.time() - os.path.getmtime(file_path)) / (24 * 3600)
+            if file_age < _TTL_DAYS:
+                logging.info(
+                    f"Skipping already downloaded file (age: {file_age:.1f} days): {file_name_to_save}"
+                )
+                continue
+            logging.info(
```
ajaits, line 1297: Also delete the file so the ttl check can be avoided next time?

Or even create a cleanup() function that deletes all old files in gcs_folder, then you don't need to check ttl for each file and also old files that are renamed will get deleted.
o/w keeping very old files in GCS cached folder could slow future runs that need to copy the files to local folder.

ajaits (approved):

scripts/us_census/pep/us_pep_sex/process.py
niveditasing: Done

scripts/us_census/pep/us_pep_sex/process.py
niveditasing: changed to gcs_folder

scripts/us_census/pep/us_pep_sex/process.py
niveditasing: I've separated the cleanup logic into a dedicated cleanup() function that scans the cache directory and deletes any file older than _TTL_DAYS, regardless of whether it's in the current download list or not.

PR #1958, opened 2026-04-16
 Rename local SDMX support package and remove facet extraction from metadata extractor

rohitkumarbhagat:
This change removes fragile facet handling from the SDMX metadata extractor and renames the local agentic-import sdmx support folder  to avoid confusion with the third-party sdmx package.


ajaits (approved):
Thanks for the quick fix!

PR #1964, opened 2026-04-17
V0 version of embedding ingestion core.

shixiao-coder:
It includes:
requirements.txt to install the related packages
placeholder for main.py function: placeholder to trigger various embedding ingestion logic
embedding_utils.py including function to:
 - read the latest timestamp for the lock, if no timestamp set None
 - function to find Node IDs to update based on timestamp, nodetype, validity by none empty
 - function to convert the Node and fields to ID and embedding_content
 - function to generate embeddings from the ID and embedding content in batch


import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    logging.info(f"Generating embeddings for {len(nodes)} nodes in batches of {BATCH_SIZE}.")
+
+    sql = """
+        INSERT OR UPDATE INTO NodeEmbeddings (subject_id, embedding_content, embeddings)
+        SELECT subject_id, content, embeddings.values
+        FROM ML.PREDICT(
+            MODEL text_embeddings,
+            (SELECT subject_id, embedding_content AS content, "RETRIEVAL_QUERY" AS task_type FROM UNNEST(@nodes))
+        )
+    """
```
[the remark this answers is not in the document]
shixiao-coder: INSERT OR UPDATE can be used and validated

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    
+    nodes = []
+    try:
+        with database.snapshot() as snapshot:
+            results = snapshot.execute_sql(sql, params=params, param_types=param_types)
+            fields = None
+            for row in results:
+                if fields is None:
+                    fields = [field.name for field in results.fields]
+                nodes.append(dict(zip(fields, row)))
```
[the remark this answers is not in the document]
shixiao-coder: tested, and this is not the case for the spanner sql output

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,169 @@
...
+
+"""Helper utilities for embedding workflows."""
+
+import logging
+import time
+from datetime import datetime
+from google.cloud.spanner_v1.param_types import TIMESTAMP, STRING, Array, Struct, StructField
+
+
+def get_latest_lock_timestamp(database):
```
gmechali, line 24: Why do you need this? If this job runs within the ingestion workflow, it will need to be in between the lock acquisition stage, and lock release stage (first and last stage of the workflow).

So you would be guaranteed to have the lock. Is this just to safeguard against a potential one-off run?

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,166 @@
...
+        logging.info(f"Filtering valid nodes updated after {timestamp}")
+        params["timestamp"] = timestamp
+        param_types["timestamp"] = TIMESTAMP
+    else:
+        logging.info("No timestamp provided, reading all valid nodes.")
+    
+    nodes = []
+    try:
+        with database.snapshot() as snapshot:
+            results = snapshot.execute_sql(updated_node_sql, params=params, param_types=param_types)
```
gmechali, line 79: I dont think this is going to work. This is very likely going to hit some major memory issues. Keep in mind that some imports can have billions of nodes, so using this python script here imo is not going to be scalable for our needs.

gmechali: Have you run it at the scale of our largest import?

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: This is to extract the time_threshold to find Node to update for incremental update

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: just updated passing stream instead of save to memory to get rid of OOM issue.

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2026 Google LLC
```
gmechali, line 1: Make sure to add this to every file!

import-automation/workflow/embedding-helper/embedding_utils.py
gmechali: Ah i see, thanks for clarifying!

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    
+    Args:
+        database: google.cloud.spanner.Database object.
+        timestamp: datetime object to filter by.
+        node_types: A list of strings representing the node types to filter by.
+        
+    Yields:
+        Dictionaries containing subject_id and name.
+    """
+    timestamp_condition = "update_timestamp > @timestamp" if timestamp else "TRUE"
```
gmechali, line 58: Just to double check, did you get approval from @keyurva  and the data team to make this change to the schema to support the timestamp on the Node Table?

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    
+    updated_node_sql = f"""
+        SELECT subject_id, name, types FROM Node 
+        WHERE name IS NOT NULL
+          AND {timestamp_condition}
+          AND EXISTS (
+            SELECT 1 FROM UNNEST(types) AS t WHERE t IN UNNEST(@node_types)
+          )
```
gmechali, line 66: I was looking at this query for the Node table. Since we’re at 160M rows and growing, I’m worried that using EXISTS(UNNEST...) without a specialized index will force a full table scan. At our current scale, that’s going to get expensive and slow. Should we look into implementing a secondary Index on the UNNEST to keep this efficient?

CREATE INDEX idx_node_types_unnest 
ON Node(UNNEST(types)) 
STORING (name, subject_id);


updated_node_sql = f"""
    SELECT subject_id, name, types 
    FROM Node 
    WHERE name IS NOT NULL
      AND {timestamp_condition}
      AND EXISTS (
        SELECT 1 
        FROM UNNEST(Node.types) AS t 
        INNER JOIN UNNEST(@node_types) AS filter_type ON t = filter_type
      )
"""

Might need to think through additional storage cost and if it makes this worth it.
Do you know how long the query takes today?

gmechali (commented):
Thanks Xiao, this looks great, left a coupe minor comments + one on the query and maybe how to optimize it.

My main feedback is that we need tests to go in this PR too! Can you add a tests/ directory and add unit tests for embedding_utils and e2e tests for main.py?

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: I tested similar Query for 1 Billion Node, this takes 30s to run and we should be fine on this effeciency

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: sure, added

shixiao-coder:
> Thanks Xiao, this looks great, left a coupe minor comments + one on the query and maybe how to optimize it.
> 
> My main feedback is that we need tests to go in this PR too! Can you add a tests/ directory and add unit tests for embedding_utils and e2e tests for main.py?

Added @gmechali

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: discussed in meeting and will add the new timestamp field into Node Table

import-automation/workflow/embedding-helper/test/embedding_utils_test.py
```diff
@@ -0,0 +1,111 @@
...
+# limitations under the License.
+
+import unittest
+from unittest.mock import MagicMock, patch
+from datetime import datetime
+import sys
+import os
+
+# Add parent directory of current file (src directory) to the path
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
gmechali: what;s this doing? and should the import line below be grouped with the ones above?

import-automation/workflow/embedding-helper/test/main_test.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @patch.dict(os.environ, {
+        "SPANNER_INSTANCE": "test-instance",
+        "SPANNER_DATABASE": "test-db",
+        "SPANNER_PROJECT": "test-proj"
+    })
+    @patch('main.spanner.Client')
+    @patch('main.get_latest_lock_timestamp')
+    @patch('main.get_updated_nodes')
+    @patch('main.filter_and_convert_nodes')
```
gmechali: This test doesnt seem super comprehensive here.

The first thing is that you dont need to mock filter_and_convert_nodes, that code can execute in the test!
The other ones (get_latest_lock, get_updated_nodes, and filter_and_convert_nodes are mocked because of the spanner reliance, but instead of mocking the whole function, you could just mock the spanner calls.

Have mock data be returned by the executeSQL calls, but otherwise have the functions actually run. It would make for a much stronger test!

import-automation/workflow/embedding-helper/embedding_utils.py
```diff
@@ -0,0 +1,169 @@
...
+import itertools
+import logging
+import time
+from datetime import datetime
+from google.cloud.spanner_v1.param_types import TIMESTAMP, STRING, Array, Struct, StructField
+
+
+_BATCH_SIZE = 500
+
+def get_latest_lock_timestamp(database):
```
vish-cs, line 26: Would it be better to move this logic into the ingestion_helper/spanner_client.py so metadata handling logic can be in a single place?

import-automation/workflow/embedding-helper/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.10-slim
```
vish-cs: Curious why we have this docker file? We want to deploy this as a cloud run server instead of a cloud function?

import-automation/workflow/embedding-helper/embedding_utils.py
shixiao-coder: yep, I'm planning to move it there in a following PR. lmk if it would be better to just move it there in the same PR

import-automation/workflow/embedding-helper/Dockerfile
shixiao-coder: The runtime is a bit long with 50-60min and processing embedding data is more like actual job and might be a bit heavy for function.

I'm open to switch to function if needed.

import-automation/workflow/embedding-helper/test/embedding_utils_test.py
shixiao-coder: Moving to use __init__.py to bundle the test folder together into the same project instead

import-automation/workflow/embedding-helper/test/main_test.py
shixiao-coder: Thanks for capturing this, just updated the mocks to just mock the SQL part with return values, and also validate various input output relating to it

import-automation/workflow/embedding-helper/Dockerfile
gmechali: Xiao, I thought the plan was to use the ingestion helper, right? (with possibility of using the aggreation helper). Either way we dont need a new dockerfile? Are you just using it to test it right now?

I dont think the duration of the run being ~1 hour makes it too heavy for a function. These things can scale! Regardless, ingestion helper is a Cloud run Service, and since you will be running this within the lock acquisition workflow, the ingestion helper will be essentially all yours resrouce-wise.

gmechali (approved):
LGTM - I think im with vishal that you should get rid of the Dockerfile (since I expect your next PR is to move it to ingestion helper and productionize it within it!)

Lastly, can you add a README.md file here? 
Approving now, but pls address these before :)

import-automation/workflow/embedding-helper/Dockerfile
shixiao-coder: dockerfile removed

vish-cs: Functions have a 1 hour timeout but a cloud run service makes more sense for this use-case.

import-automation/workflow/embedding-helper/embedding_utils.py
vish-cs: We should only move the get_latest_lock_timestamp in the ingestion helper. Rest of the logic is embedding specific so that can stay separate.

vish-cs (approved):

PR #1969, opened 2026-04-20
Fixes for import automation

vish-cs:
* Fix the condition to check for an empty diff (schema_diff_size)
* Add the option to import-helper to run the batch job or the dataflow job
* Update spanner client to only return staging imports

### DIFF
--- a/import-automation/executor/app/executor/import_executor.py
+++ b/import-automation/executor/app/executor/import_executor.py
@@ -548,16 +548,15 @@ class ImportExecutor:
         return base_config_path
 
     @log_function_call
-    def _invoke_import_differ(self, genmcf_output_path: str,
-                              validation_output_path: str, import_name: str,
-                              import_prefix: str, version: str,
-                              latest_version: str, import_input: dict,
-                              absolute_import_dir: str) -> Tuple[str, str]:
+    def _invoke_differ_summary(self, genmcf_output_path: str,
+                               validation_output_path: str, import_name: str,
+                               import_prefix: str, version: str,
+                               latest_version: str, import_input: dict,
+                               absolute_import_dir: str) -> Optional[Dict]:
         """Invokes the differ tool to compare current data with previous data."""
         current_data_path = os.path.join(genmcf_output_path, '*.mcf')
         previous_data_path = os.path.join(latest_version, import_prefix,
                                           'genmcf', '*.mcf')
-        diff_found = True
         # TODO: remove fallback logic once all imports move to new path.
         if not file_util.file_get_matching(previous_data_path):
             input_files = self._get_import_input_files(import_input,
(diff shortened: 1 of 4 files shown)


import-automation/executor/app/executor/import_executor.py
```diff
@@ -716,7 +716,15 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
...
-        return validation_status, differ_status
+        if not self.config.ignore_validation_status and not validation_status:
+            logging.error(
+                "Marking import as VALIDATION due to validation failure.")
+            import_summary.status = ImportStatus.VALIDATION
+        elif not differ_status:
+            logging.info("Marking import as SKIP due to no data diff.")
+            import_summary.status = ImportStatus.SKIP
+        else:
+            import_summary.status = ImportStatus.STAGING
```
ajaits, line 728: `return validation_status`

import-automation/workflow/import-helper/main.py
```diff
@@ -43,8 +43,21 @@ def handle_feed_event(request):
     latest_version = attributes.get(
         'import_version',
         datetime.now(timezone.utc).strftime("%Y-%m-%d"))
-    run_ingestion = True 
+    post_process = attributes.get('post_process', '')
```
ajaits: Is this a new attribute? How is this used?

The name seems to indicate this is post running import workflow.
Can we rename this to `run_process` so when set to `spanner_ingestion_workflow` it is clear that is it only running a dataflow ingestion?

ajaits (approved):

import-automation/executor/app/executor/import_executor.py
```diff
@@ -656,6 +652,10 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
                     latest_version=latest_version,
                     import_input=import_input,
                     absolute_import_dir=absolute_import_dir)
+                if differ_summary is not None:
+                    diff_found = (differ_summary['obs_diff_size'] != 0 or
```
ajaits: can we use `.get()` instead of `[]`?
`diff_summary.get('obs_diff_size', 0) != 0 or differ_summary,get('schema_diff_size', 0) != 0 `

vish-cs: We should not use 0 as the default value as that can lead to the import being marked as SKIP. I have changed the default value to 1

import-automation/executor/app/executor/import_executor.py
vish-cs: Done

import-automation/workflow/import-helper/main.py
vish-cs: These are just key-word pairs that we can attach. The idea is to be able to configure which imports should first run the batch job and which imports should directly run the dataflow job. I have renamed.

PR #1971, opened 2026-04-20
[DCP Ingestion] Automate the DB initialization, Dockerize Ingestion Helper, and fix lock acquisition issue

gmechali:
This PR does 3 main things:
- It dockerizes the ingestion helper such that we can run it in a Cloud Run Server. Today for base DC we deploy it as a Cloud Run Function, but that isnt ideal for DCP. For DCP, we will deploy a Cloud run server on the datacommon-ingestion-helper image. I have added a [cloud build trigger](https://pantheon.corp.google.com/cloud-build/triggers;region=global/edit/750fd9da-466a-43c2-80d1-e9479eed0e4c?e=13803378&inv=1&invt=Ab3CEA&mods=-monitoring_api_staging&project=datcom-ci) in datcom-ci to rebuild the image everytime we make a change to the master branch on the ingestion-helper directory.
- Adds the Initialize Database action to the IngestionHelper function, which initializes all the tables in spanner_schema.sql (Edge, Node, Observation, ImportStatus, IngestionLock, ImportVersionHistory, IngestionHistory). 
- Modifies the Lock Acquisition step to insert the lock row for the first time this is run.

As a follow up we should do the following:
- Migrate the Base DC deployed cloud run function to use this dockerized imaged and deploy it in cloud run server.
- Remove all the logic from the ingestion pipeline that initializes the database.

### DIFF
--- a/import-automation/workflow/ingestion-helper/Dockerfile
+++ b/import-automation/workflow/ingestion-helper/Dockerfile
@@ -25,12 +25,11 @@ WORKDIR /app
 # Install protobuf compiler and curl
 RUN apt-get update && apt-get install -y protobuf-compiler curl && rm -rf /var/lib/apt/lists/*
 
-# Install production dependencies.
-COPY requirements.txt .
-RUN uv pip install --system --no-cache -r requirements.txt
-
 # Copy local code to the container image.
 COPY . .
+
+# Install production dependencies using uv.
+RUN uv pip install --system --no-cache .
 
 # Fetch proto file from GitHub
 RUN curl -o storage.proto https://raw.githubusercontent.com/datacommonsorg/import/master/pipeline/data/src/main/proto/storage.proto
--- a/import-automation/workflow/ingestion-helper/Dockerfile
+++ b/import-automation/workflow/ingestion-helper/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM python:3.12-slim
+
+# Allow statements and log messages to immediately appear in the logs
+ENV PYTHONUNBUFFERED True
+
+WORKDIR /app
+
+# Install protobuf compiler and curl
+RUN apt-get update && apt-get install -y protobuf-compiler curl && rm -rf /var/lib/apt/lists/*
+
+# Install production dependencies.
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Copy local code to the container image.
+COPY . .
+
+# Fetch proto file from GitHub
+RUN curl -o storage.proto https://raw.githubusercontent.com/datacommonsorg/import/master/pipeline/data/src/main/proto/storage.proto
+
+# Generate proto descriptor set
+RUN protoc --include_imports --descriptor_set_out=storage.pb storage.proto
+
+# Run the functions framework
+CMD ["functions-framework", "--target", "ingestion_helper"]
--- a/import-automation/workflow/ingestion-helper/Dockerfile
+++ b/import-automation/workflow/ingestion-helper/Dockerfile
@@ -14,6 +14,9 @@
 
 FROM python:3.12-slim
 
+# Copy uv binary
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
 # Allow statements and log messages to immediately appear in the logs
 ENV PYTHONUNBUFFERED True
 
@@ -24,7 +27,7 @@ RUN apt-get update && apt-get install -y protobuf-compiler curl && rm -rf /var/l
 
 # Install production dependencies.
 COPY requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt
+RUN uv pip install --system --no-cache -r requirements.txt
 
 # Copy local code to the container image.
 COPY . .
--- a/import-automation/workflow/ingestion-helper/README.md
+++ b/import-automation/workflow/ingestion-helper/README.md
@@ -56,3 +56,8 @@ Updates the version of an import, records version history, and updates the statu
 *   `version` (Required): The version string. If set to `'STAGING'`, it resolves to the current staging version.
 *   `comment` (Required): A comment for the audit log explaining the version update.
 *   `override` (Optional): Override version without checking import status (boolean)
+
+#### `initialize_database`
+Initializes the Spanner database by creating all necessary tables and uploading proto descriptors.
+
+*   This action requires no payload parameters. It automatically reads `schema.sql` and `storage.pb` from the container directory to provision the database schema and proto descriptors.
--- a/import-automation/workflow/ingestion-helper/README.md
+++ b/import-automation/workflow/ingestion-helper/README.md
@@ -61,3 +61,24 @@ Updates the version of an import, records version history, and updates the statu
 Initializes the Spanner database by creating all necessary tables and uploading proto descriptors.
 
 *   This action requires no payload parameters. It automatically reads `schema.sql` and `storage.pb` from the container directory to provision the database schema and proto descriptors.
+*   **Note on Protos**: The `storage.pb` file is generated during the Docker build process. The `Dockerfile` fetches `storage.proto` from the `datacommonsorg/import` GitHub repository and compiles it into `storage.pb`.
+
+## Local Development and Testing
+
+To run the helper service locally and test its functionality:
+
+### Running the Server
+Ensure you have installed the requirements (`uv pip install -r requirements.txt`), then start the functions framework:
+
+```bash
+uv run functions-framework --target ingestion_helper
+```
+By default, this will start serving on `http://localhost:8080`.
+
+### Triggering Actions
+You can test specific actions by sending a POST request with a JSON payload. For example, to trigger database initialization locally:
+```bash
+curl -X POST http://localhost:8080 \
+  -H "Content-Type: application/json" \
+  -d '{"actionType": "initialize_database"}'
+```
--- a/import-automation/workflow/ingestion-helper/main.py
+++ b/import-automation/workflow/ingestion-helper/main.py
@@ -193,5 +193,10 @@ def ingestion_helper(request):
             f"OK [Import: {import_name} Version: {version} Status: {params['status']}]",
             200)
 
+    elif actionType == 'initialize_database':
+        # Initializes the database by creating all required tables and proto bundles.
+        logging.info("Action: initialize_database")
+        spanner.initialize_database()
+        return ('OK', 200)
     else:
         return (f'Unknown actionType: {actionType}', 400)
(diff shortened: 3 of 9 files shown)


import-automation/workflow/ingestion-helper/spanner_client.py
```diff
@@ -383,3 +398,67 @@ def _record(transaction: Transaction):
...
+                database=database_path,
+                statements=ddl_statements,
+                proto_descriptors=proto_descriptors_bytes
+            )
+            operation = admin_client.update_database_ddl(request=request)
+            operation.result()
+            logging.info("Database initialized successfully with protos.")
+        except Exception as e:
+            logging.error(f"Failed to update DDL with protos: {e}")
+            raise
```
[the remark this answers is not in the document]
gmechali: It's all deployed in the same account so not needed.

import-automation/workflow/ingestion-helper/schema.sql
```diff
@@ -0,0 +1,110 @@
+-- Copyright 2026 Google LLC
```
gmechali, line 1: This file shouldbe the source of truth for Database initialization. 

Note that today's it s split across:
- Graph information: https://github.com/datacommonsorg/import/blob/591b7df8d1a85d0eab1e207448dcc6b344c2c87e/pipeline/spanner/src/main/resources/spanner_schema.sql#L4
- Import information: https://github.com/datacommonsorg/data/blob/cddc0a7a5c5f29bdf98b319c7dae3f66338ff653/import-automation/workflow/spanner_schema.sql#L4

As a follow up, once we've moved our base DC infra to use this new cloud run server, we should delete both of those schema files above.

import-automation/workflow/ingestion-helper/main.py
```diff
@@ -193,5 +193,10 @@ def ingestion_helper(request):
             f"OK [Import: {import_name} Version: {version} Status: {params['status']}]",
             200)
 
+    elif actionType == 'initialize_database':
```
vish-cs, line 196: Please also update the README to add the new action.

import-automation/workflow/ingestion-helper/schema.sql
```diff
@@ -0,0 +1,110 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+CREATE PROTO BUNDLE (
+  `org.datacommons.Observations`
+);
+
+CREATE TABLE Node (
```
vish-cs, line 19: If we are moving the schema here, we should remove the existing schema files from below to avoid inconsistency. This will require some changes since the existing Terraform script and integration tests use these files.
https://github.com/datacommonsorg/import/blob/master/pipeline/spanner/src/main/resources/spanner_schema.sql
and
https://github.com/datacommonsorg/data/blob/master/import-automation/workflow/spanner_schema.sql

import-automation/workflow/ingestion-helper/main.py
gmechali: Ah yes good catch. Done

import-automation/workflow/ingestion-helper/schema.sql
gmechali: yes correct but doing that is also going to need us to replace our own base DC Cloud Function to a server. 
Im happy to do this as a follow up but wanted to make sure you / Dave were onboard.

import-automation/workflow/ingestion-helper/spanner_client.py
```diff
@@ -383,3 +398,67 @@ def _record(transaction: Transaction):
...
+        logging.info(f"Reading schema from {schema_path}")
+        try:
+            with open(schema_path, 'r') as f:
+                schema_content = f.read()
+            ddl_statements = [s.strip() for s in schema_content.split(';') if s.strip()]
+        except Exception as e:
+            logging.error(f"Failed to read schema file: {e}")
+            raise
+            
+        proto_path = os.path.join(os.path.dirname(__file__), 'storage.pb')
```
dwnoble, line 439: Can you add a comment describing the proto & where it's fetched from?

import-automation/workflow/ingestion-helper/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY . .
+
+# Fetch proto file from GitHub
+RUN curl -o storage.proto https://raw.githubusercontent.com/datacommonsorg/import/master/pipeline/data/src/main/proto/storage.proto
+
+# Generate proto descriptor set
+RUN protoc --include_imports --descriptor_set_out=storage.pb storage.proto
+
+# Run the functions framework
+CMD ["functions-framework", "--target", "ingestion_helper"]
```
dwnoble, line 39: I don't think the gcp functions-framework utility is the right tool to use here for hosting a rest service. 

Is the plan to migrate all of these functions to cloud run services?

If so- i think it makes sense to put them in the DCP service as a new endpoint:

https://github.com/datacommonsorg/datacommons/tree/main/packages/datacommons-api/datacommons_api/

This way we wont need a new image or ci/cd pipeline for building the import automation docker image.

Also- moving from functions-framework to fastapi gives us better logging, request & response validations with Pydantic objects, and puts these APIs in a single place.

import-automation/workflow/ingestion-helper/README.md
```diff
@@ -56,3 +56,8 @@ Updates the version of an import, records version history, and updates the statu
 *   `version` (Required): The version string. If set to `'STAGING'`, it resolves to the current staging version.
 *   `comment` (Required): A comment for the audit log explaining the version update.
 *   `override` (Optional): Override version without checking import status (boolean)
+
+#### `initialize_database`
+Initializes the Spanner database by creating all necessary tables and uploading proto descriptors.
+
+*   This action requires no payload parameters. It automatically reads `schema.sql` and `storage.pb` from the container directory to provision the database schema and proto descriptors.
```
dwnoble, line 63: Can you expand on where storage.proto / storage.pb is pulled from?

Also-can you add an example of how to run the command locally?

dwnoble (commented):
Thanks Gabe!

I think this tooling may be better suited to live in the DCP repo at https://github.com/datacommonsorg/datacommons

This PR converts the import automation cloud functions to a REST service, but if we're going to do that, I think it makes sense to consolidate it with the DCP FastAPI service so we don't end up adding a separate process.

Then this action could look something like: `POST https://<dcp-service>/api/admin/initialize-database`

Regardless, i think we should also:
- convert this toolset from requirements.txt files to uv so we can have consistent builds with the uv.lock files
- right now the functions in main.py are tied to cloud run functions. we should migrate these to a plain python module, and the cloud run part should be minimal wrapper on top of it
- we should expose these functions via a command line tool. If we migrate to the dcp repo, this might look like: `$ datacommons admin initialize-database` or `$ datacommons admin release-ingestion-lock`.

cc @vish-cs

import-automation/workflow/ingestion-helper/README.md
gmechali: This is in the dockerfile:

RUN curl -o storage.proto https://raw.githubusercontent.com/datacommonsorg/import/master/pipeline/data/src/main/proto/storage.proto

RUN protoc --include_imports --descriptor_set_out=storage.pb storage.proto

Added instructions for running locally.

import-automation/workflow/ingestion-helper/spanner_client.py
gmechali: It's downloaded in the Dockerfile from Github and runs protoc. So  when this runs it's already in the container. Im not sure what's worth adding in the comment? Just that we set it up in the container through the Dockerfile?

import-automation/workflow/ingestion-helper/Dockerfile
gmechali: I agree that functions-framework is a bit limited and odd in this case, but this is essentially the minimum work to migrate the existing cloud run function to a cloud run service executing the cloud run function. It allows dockerization for DCP, without bloating Base DC infra/

I think there needs to be much more discussion before moving this to DCP API, it certainly has some pros but also some major drawbacks such as moving away ingestion code used across base DC in this new DCP repo. It might be premature to move to DCP service, especially as DCP service is out of scope forM1. 

If we leave it like this M1 ships 4 artifacts:
- MCP PyPi package
- Mixer image
- ingestion helper image
- dataflow template
all connected through the Terraform deployment. If we move ingestion helper to DCP service, then we just need to maintain the DCP service image. Agreed functions framework is limited and not ideal, I think that migration can happen at any point in the future. There is no need to do it now, and there would be no complexity in migrating it to within DCP service when we launch DCP service (in q3/q4). 
Especially as ingestion helper is required for base DC, I dont think it makes sense for us to take this on now and we would have to deploy DCP service to Base DC for ingestion helper support. 

Happy to hop on a call to discuss furhter!

gmechali:
> Thanks Gabe!
> 
> I think this tooling may be better suited to live in the DCP repo at https://github.com/datacommonsorg/datacommons
> 
> This PR converts the import automation cloud functions to a REST service, but if we're going to do that, I think it makes sense to consolidate it with the DCP FastAPI service so we don't end up adding a separate process.
> 
> Then this action could look something like: `POST https://<dcp-service>/api/admin/initialize-database`
> 
> Regardless, i think we should also:
> 
> * convert this toolset from requirements.txt files to uv so we can have consistent builds with the uv.lock files
> * right now the functions in main.py are tied to cloud run functions. we should migrate these to a plain python module, and the cloud run part should be minimal wrapper on top of it
> * we should expose these functions via a command line tool. If we migrate to the dcp repo, this might look like: `$ datacommons admin initialize-database` or `$ datacommons admin release-ingestion-lock`.
> 
> cc @vish-cs

Other comment kind of replies about moving to DCP repo. I think one day maybe but it's premature, and splitting up ingestion code also seems controversial. 
Converting to UV - done!
Migrating from cloud functions to a plain python module, agreed but I dont think there's an immediate need for it? Can we track this post-M1?
Command line tool, again I think this is way past M1 right? All nice to haves, but are we aligned that these are future improvements, not for MVP?

import-automation/workflow/ingestion-helper/README.md
dwnoble: sg!

import-automation/workflow/ingestion-helper/Dockerfile
dwnoble: This makes sense given our timeline, but i would like to prioritize migrating the capability to the dcp service post M1.

i do think it makes sense to run a DCP service in our base DC environment sooner rather than later, and this seems like a good use case for it.

dwnoble (approved):
> Other comment kind of replies about moving to DCP repo. I think one day maybe but it's premature, and splitting up ingestion code also seems controversial.
Converting to UV - done!
Migrating from cloud functions to a plain python module, agreed but I dont think there's an immediate need for it? Can we track this post-M1?
Command line tool, again I think this is way past M1 right? All nice to haves, but are we aligned that these are future improvements, not for MVP?

Adding the python module & CLI later SGTM.

For converting to uv, we should also migrate from requirements.txt to pyproject.toml (example: https://github.com/datacommonsorg/datacommons/blob/main/packages/datacommons-db/pyproject.toml)

but that can also happen in a follow-up!

dhotchkiss (approved):

gmechali:
> > Other comment kind of replies about moving to DCP repo. I think one day maybe but it's premature, and splitting up ingestion code also seems controversial.
> > Converting to UV - done!
> > Migrating from cloud functions to a plain python module, agreed but I dont think there's an immediate need for it? Can we track this post-M1?
> > Command line tool, again I think this is way past M1 right? All nice to haves, but are we aligned that these are future improvements, not for MVP?
> 
> Adding the python module & CLI later SGTM.
> 
> For converting to uv, we should also migrate from requirements.txt to pyproject.toml (example: https://github.com/datacommonsorg/datacommons/blob/main/packages/datacommons-db/pyproject.toml)
> 
> but that can also happen in a follow-up!

Changed to pyproject.toml!

PR #1975, opened 2026-04-22
Use import tool report for EMPTY_IMPORT_CHECK

vish-cs:


import-automation/workflow/import-helper/main.py
```diff
@@ -48,7 +48,7 @@ def handle_feed_event(request):
     if import_step == 'spanner_ingestion_workflow':
         # Update import status in spanner
         import_status = 'STAGING'
-        graph_path = attributes.get('graph_path', "/**/*.*")
+        graph_path = attributes.get('graph_path', "/**/*.mcf*")
```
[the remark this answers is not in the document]
vish-cs: This key is different from the key in the manifest.

scripts/entities/download.sh
```diff
@@ -37,7 +37,7 @@ GCS_PATH="gs://${BUCKET_NAME}/${GCS_FOLDER_PREFIX}/${VERSION}"
 
 echo "Downloading import ${ENTITY} for version ${VERSION} from ${GCS_PATH} to $(pwd)"
 mkdir -p "${ENTITY}"
-gcloud storage cp -r "${GCS_PATH}" "${ENTITY}/"
+gcloud storage cp -r "${GCS_PATH}" "${ENTITY}/" &> copy.log
```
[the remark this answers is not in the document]
vish-cs: This is required to mask all copy log statements which can overwhelm logging.

tools/import_validation/validator.py
ajaits: Wouldn't summary be None for an import that has never been run (or run successfully) where there is no previous successful run to diff against?

tools/import_validation/validator.py
```diff
@@ -300,6 +300,14 @@ def validate_empty_import(self, differ_df: pd.DataFrame, summary: dict,
                                     'EMPTY_IMPORT_CHECK',
                                     message="Differ summary is missing.")
 
+        if 'current_obs_size' not in summary or 'current_schema_size' not in summary:
```
ajaits: Can we use the validation report.json instead of differ for checking empty import output or even just check the import output files rows for imports that may have disabled differ and validations?

vish-cs: I have udpated the check to use report.json counters

tools/import_validation/validator.py
vish-cs: I have udpated the check to use report.json counters

ajaits (approved):

PR #1994, opened 2026-05-06
Migrating the embedding helper functionality into ingestion helper

shixiao-coder:
This PR moves the embedding_utils.py into ingestion-helper and add a new actionType embedding_ingestion to use the related fuctionality.
We use the flag enable_embedding to control if we actually trigger embeddings and a flag nodetype that default to SVs+Topics to control which nodes to embed

Tests are also migrated into ingestion-helper


gmechali (approved):
so i assume this enableEmbeddings check is for you to be able to add another control, and ensure that we dont accidentally trigger embeddings before we want it launched.

I think that s a great addition! Can you just return an error code, not OK because that is now an invalid request here.

shixiao-coder:
> so i assume this enableEmbeddings check is for you to be able to add another control, and ensure that we dont accidentally trigger embeddings before we want it launched.
> 
> I think that s a great addition! Can you just return an error code, not OK because that is now an invalid request here.

Updated

PR #1998, opened 2026-05-11
Update differ to ouptut MCF files

vish-cs:


tools/import_validation/runner.py
```diff
@@ -167,6 +144,51 @@ def _initialize_data_sources(self, stats_summary: str, lint_report: str,
             logging.warning("lint_report file exists but is empty: %s",
                             lint_report)
 
+    def _load_differ_df_from_mcf(self, input_dir: str) -> pd.DataFrame:
+        """Parses MCF diff files and returns a summary DataFrame."""
+        import glob
+        from collections import defaultdict
+        stats = defaultdict(lambda: {'ADDED': 0, 'DELETED': 0, 'MODIFIED': 0})
+
+        combined_pattern = os.path.join(input_dir, 'import-diff-*.mcf')
```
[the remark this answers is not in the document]
vish-cs: Updated the pattern

tools/import_differ/import_differ.py
```diff
@@ -65,7 +66,7 @@
   'Path (local/GCS) to the output data folder.')
 flags.DEFINE_string('file_format', 'mcf',
                     'Format of the input data (mcf,tfrecord)')
-flags.DEFINE_string('runner_mode', 'local', 'Runner mode (local/cloud)')
+flags.DEFINE_string('runner_mode', 'local', 'Runner mode (local/cloud/direct)')
```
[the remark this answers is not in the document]
vish-cs: Using direct for Java direct runner and native/local for Python

tools/import_validation/runner.py
```diff
@@ -121,38 +123,13 @@ def _initialize_data_sources(self, stats_summary: str, lint_report: str,
...
-                logging.error(
-                    f"JSON parse error while reading differ summary at {differ_json_path}: {e}"
-                )
-        elif differ_json_path and os.path.exists(differ_json_path):
-            logging.warning("differ summary file exists but is empty: %s",
-                            differ_json_path)
+            differ_json_path = os.path.join(differ_output,
+                                            'differ_summary.json')
+            with open(differ_json_path, 'r') as f:
+                self.data_sources['differ_summary'] = json.load(f)
```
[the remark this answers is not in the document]
vish-cs: Added back

tools/import_differ/differ_utils.py
```diff
@@ -84,6 +84,28 @@ def write_json_data(data, dest: str, file: str, tmp_dir: str):
         upload_output_data(path, dest)
 
 
+def write_mcf_nodes(nodes: list, dest: str, file: str, tmp_dir: str):
+    """ Writes mcf nodes to a file with the given path."""
+    if dest.startswith('gs://'):
+        path = os.path.join(tmp_dir, file)
+    else:
+        path = os.path.join(dest, file)
+    with open(path, mode='w', encoding='utf-8') as out_file:
```
ajaits: You can use util/file_util.py:FileIO() to support GCS or local files transparently without needing extra GCS file handling code here:
```
with FileIO(path, mode='w', encoding='utf-8') as out_file:
...
```

tools/import_differ/import_differ.py
```diff
@@ -65,7 +66,8 @@
   'Path (local/GCS) to the output data folder.')
 flags.DEFINE_string('file_format', 'mcf',
                     'Format of the input data (mcf,tfrecord)')
-flags.DEFINE_string('runner_mode', 'local', 'Runner mode (local/cloud)')
+flags.DEFINE_string('runner_mode', 'native',
```
ajaits, line 74: Can we add a brief description of the different modes in the file? The difference between native and direct is not clear by the  term.

tools/import_differ/import_differ.py
```diff
@@ -231,72 +224,75 @@ def split_data(self, mcf_nodes: list) -> (pd.DataFrame, pd.DataFrame):
...
-        summary = result.set_index(
-            Column.diff_type.name).transpose().rename_axis(
-                None, axis=1).reset_index(drop=True)
-        self._cleanup_data(summary)
-        return summary
+        all_nodes = []
+        for diff_type in [
+                Diff.ADDED.name, Diff.DELETED.name, Diff.MODIFIED.name
+        ]:
+            df_type = diff_df[diff_df[Column.diff_type.name] == diff_type]
```
ajaits: If the obs nodes are being filtered separately by type, can we write them into separate mcf files too? 
That may help with other analysis and also skip the `diffType` property in the node.

tools/import_differ/import_differ.py
```diff
@@ -231,72 +224,75 @@ def split_data(self, mcf_nodes: list) -> (pd.DataFrame, pd.DataFrame):
...
+                key_combined = str(row[Column.key_combined.name])
+
+                # Determine which column to use for values and node IDs
+                suffix = '_x' if diff_type == Diff.DELETED.name else '_y'
+
+                # Helper to get value from row, handles cases with or without suffix
+                def get_val(base_name):
+                    col_name = base_name + suffix
+                    if col_name in row:
+                        return str(row[col_name])
```
ajaits, line 270: Instead of `str()`, can we have a utility to convert it to a string?
for example, for lists, instead of `[... ]`, MCFs use `a,b,c`. 
Also if the value has spaces, it needs to be enclosed in double quotes.

```
def val_str(value) -> str:
    if isinstance(value, list):
        return ",".join([val_str(v) for v in value])
    if value and isinstance(value, str) and " " in value and value[0].isalpha():
        return '"' + value + '"'
    return str(value)
```

tools/import_differ/import_differ.py
ajaits: Can we make this a constant since it is used in multiple places?

tools/import_differ/differ_utils.py
vish-cs: Done

tools/import_differ/import_differ.py
vish-cs: Done

tools/import_differ/import_differ.py
vish-cs: Updated to create 3 separate files nodes-added.mcf, nodes-deleted.mcf, and nodes-modified.mcf

tools/import_differ/import_differ.py
vish-cs: Done

tools/import_differ/import_differ.py
vish-cs: Done

ajaits (approved):

PR #2010, opened 2026-05-14
Added seed database ingestion helper

dwnoble:
This pull request introduces a new `seed_database` action to the ingestion helper service, which seeds the Spanner database with essential base nodes required by the Data Commons schema. The implementation includes the new action handler, the underlying logic in the Spanner client, and comprehensive tests to ensure correct behavior.

**New Feature: Database Seeding**

* Added a `seed_database` action to the ingestion helper API, which seeds the Spanner database with base nodes such as `StatisticalVariable`, `StatVarGroup`, `StatVarObservation`, `Topic`, and `c/g/Root`. This action ensures that the database contains the minimum required schema nodes for Data Commons operations. [[1]](diffhunk://#diff-eb17f9499c3ffbbfafbdd2a5f313f59f21002eb54e3b35d36a16777719f66b09R67-R71) [[2]](diffhunk://#diff-c309bb37461cb80457d43198df8293564f9b762c9cde39332465d05be19c45acR222-R226) [[3]](diffhunk://#diff-8a03a24cc5702a19efadc4f0cee2679d4c0c7cffe73397483f9bc1dc475c8f86R558-R594)

**Testing**

* Implemented unit tests in `main_test.py` to verify that the `seed_database` action is handled correctly and that the Spanner client’s `seed_database` method is called as expected.
* Added tests in `spanner_client_test.py` to ensure that the `seed_database` method inserts base nodes when missing and does not insert duplicates if the nodes already exist.

**Documentation**

* Updated the `README.md` to document the new `seed_database` action, including its purpose and usage.

### DIFF
--- a/import-automation/workflow/ingestion-helper/main.py
+++ b/import-automation/workflow/ingestion-helper/main.py
@@ -219,6 +219,11 @@ def ingestion_helper(request):
                                              FLAGS.enable_embeddings)
         spanner.initialize_database(enable_embeddings=enable_embeddings)
         return ('OK', 200)
+    elif actionType == 'seed_database':
+        # Seeds the database with base empty nodes.
+        logging.info("Action: seed_database")
+        spanner.seed_database()
+        return ('OK', 200)
     elif actionType == 'embedding_ingestion':
         logging.info("Action: embedding_ingestion")
         enable_embeddings = request_json.get('enableEmbeddings',
--- a/import-automation/workflow/ingestion-helper/README.md
+++ b/import-automation/workflow/ingestion-helper/README.md
@@ -64,6 +64,11 @@ Initializes the Spanner database by creating all necessary tables and uploading 
 *   `enableEmbeddings` (Optional): Boolean to enable creation of embedding tables and models.
 *   **Note on Protos**: The `storage.pb` file is generated during the Docker build process. The `Dockerfile` fetches `storage.proto` from the `datacommonsorg/import` GitHub repository and compiles it into `storage.pb`.
 
+#### `seed_database`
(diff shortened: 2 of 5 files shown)


import-automation/workflow/ingestion-helper/main.py
```diff
@@ -219,6 +219,11 @@ def ingestion_helper(request):
                                              FLAGS.enable_embeddings)
         spanner.initialize_database(enable_embeddings=enable_embeddings)
         return ('OK', 200)
+    elif actionType == 'seed_database':
```
gmechali: What do you think about just putting this in intitialize_database? Keep the method seed_database() but have it be triggered as the last step in initialize_datanase?

gmechali (approved):
thanks Dan, LGTM! 

TBH I'm not sure if we're missing some of the nodes required. I have a doubt on whether we need dc/g/Root - very liekly but not sure. But we can add it later if confirmed.

dwnoble:
> thanks Dan, LGTM!
> 
> TBH I'm not sure if we're missing some of the nodes required. I have a doubt on whether we need dc/g/Root - very liekly but not sure. But we can add it later if confirmed.

Thanks Gabe! And that sounds good. we can modify this list as we go

import-automation/workflow/ingestion-helper/main.py
dwnoble: I was thinking about that, but i think it's still useful to be able to call it separately (in case we need to add additional seeds for example). But, i think we can have the CLI tool call init & seed in a single operation so users won't need to make two calls

PR #2011, opened 2026-05-14
Add config to disable skip status

vish-cs:
This is required for the integration test. We repeatedly run the same import for ingestion during the test. Recently, we added a SKIP state where the import is skipped if there is no data change which happens for the integration test. This change is to force the ingestion even if there is no data change so we can verify the E2E flow.

### DIFF
--- a/import-automation/executor/app/configs.py
+++ b/import-automation/executor/app/configs.py
@@ -169,6 +169,8 @@ class ExecutorConfig:
     email_token: str = ''
     # Email alerts are disabled by default. Cloud Run jobs use GCP alerting.
     disable_email_notifications: bool = True
+    # Allow skipping ingestion for no data change.
+    enable_skip_status: bool = True
     # Skip uploading the data to GCS (for local testing).
     skip_gcs_upload: bool = False
     # Skip uploading input files to GCS.
--- a/import-automation/executor/app/executor/import_executor.py
+++ b/import-automation/executor/app/executor/import_executor.py
@@ -721,7 +721,7 @@ class ImportExecutor:
             logging.error(
                 "Marking import as VALIDATION due to validation failure.")
             import_summary.status = ImportStatus.VALIDATION
-        elif not differ_status:
+        elif self.config.enable_skip_status and not differ_status:
             logging.info("Marking import as SKIP due to no data diff.")
             import_summary.status = ImportStatus.SKIP
         else:
--- a/scripts/us_fed/treasury_constant_maturity_rates/manifest.json
+++ b/scripts/us_fed/treasury_constant_maturity_rates/manifest.json
@@ -20,7 +20,10 @@
             "source_files": [
                 "treasury_constant_maturity_rates.csv"
             ],
-            "cron_schedule": "15 3 * * *"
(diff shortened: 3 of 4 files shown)


import-automation/executor/app/executor/import_executor.py
```diff
@@ -721,7 +721,7 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
             logging.error(
                 "Marking import as VALIDATION due to validation failure.")
             import_summary.status = ImportStatus.VALIDATION
-        elif not differ_status:
+        elif self.config.enable_skip_status and not differ_status:
```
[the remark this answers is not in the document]
vish-cs: We override the global config with values in manifest.

import-automation/executor/app/configs.py
```diff
@@ -169,6 +169,8 @@ class ExecutorConfig:
     email_token: str = ''
     # Email alerts are disabled by default. Cloud Run jobs use GCP alerting.
     disable_email_notifications: bool = True
+    # Allow skipping ingestion for no data change.
+    enable_skip_status: bool = True
```
rohitkumarbhagat, line 173: consider renaming to clarity. maybe `skip_import_on_no_change`

scripts/us_fed/treasury_constant_maturity_rates/manifest.json
```diff
@@ -20,7 +20,10 @@
             "source_files": [
                 "treasury_constant_maturity_rates.csv"
             ],
-            "cron_schedule": "15 3 * * *"
+            "cron_schedule": "15 3 * * *",
+            "config_override": {
+                "enable_skip_status": false
```
rohitkumarbhagat, line 25: I assumed disable is only for test. Will this not disable during actual import.

Also, flag needs to be added to test. Is that another PR?

rohitkumarbhagat (approved):

import-automation/executor/app/executor/import_executor.py
```diff
@@ -721,7 +721,7 @@ def _invoke_import_validation(self, repo_dir: str, relative_import_dir: str,
...
             logging.info("Marking import as SKIP due to no data diff.")
```
rohitkumarbhagat, line 725: The log can be updated to indicate skip due to flag.

scripts/us_fed/treasury_constant_maturity_rates/manifest.json
vish-cs: This is a test import USFed_ConstantMaturityRates_Test using for testing only so we don't need to add a separate config for the test.

import-automation/executor/app/executor/import_executor.py
vish-cs: The flag is only to enable the SKIP state which is then enforced due to no data diff.

import-automation/executor/app/configs.py
vish-cs: Actually the flag is named to reflect that we are only changing the import status from STAGING to SKIP. We still upload the file to GCS. The downstream behavior on SKIP status is controlled by the ingestion workflow.

PR #2018, opened 2026-05-20
CensusACS5YearSurvey_SubjectTables_S2602 Config Update

saanikaaa:
PR to bring the configs changes made to run and generate output for CensusACS5YearSurvey_SubjectTables_S2602. The config file had minor changes where we extended certain values for latest years to get correct data point values and minor change in "very well" key which was breaking the processing.

PR_CHECKLIST - https://docs.google.com/spreadsheets/d/1XSBfjepI8o7SrpVAUMDcUjT7SXZ--kEtFfsTrlRoBk8/edit?resourcekey=0-cue1WQ54CzrVy2tVMQj5tQ&gid=1086229089#gid=1086229089


HarishC727:
Add a README file for the table on which Geos are included for this specific S table. Next time it becomes easy to download for only those geos

saanikaaa:
> Add a README file for the table on which Geos are included for this specific S table. Next time it becomes easy to download for only those geos

We don't limit the download. The data is not present for other levels. Once, source start bringing we can extend the same.

HarishC727:
> > Add a README file for the table on which Geos are included for this specific S table. Next time it becomes easy to download for only those geos
> 
> We don't limit the download. The data is not present for other levels. Once, source start bringing we can extend the same.

But please add a basic README file for the table

HarishC727:
LGTM

saanikaaa:
> > > Add a README file for the table on which Geos are included for this specific S table. Next time it becomes easy to download for only those geos
> > 
> > 
> > We don't limit the download. The data is not present for other levels. Once, source start bringing we can extend the same.
> 
> But please add a basic README file for the table

Basic README file for the table already exist

PR #2022, opened 2026-05-20
[DCP - BQ Federation] Propagate location in to the BQ connection

gmechali:
Allow using the env var REGION or LOCATION because base DC ingestion helper uses LOCATION, while DCP has been using the region env var everywhere. Would be great to align on one long term but this should be harmless.


clincoln8 (approved):
Thanks!!

PR #2030, opened 2026-05-22
Use asynchronous aggregation queries

vish-cs:


import-automation/workflow/ingestion-helper/aggregation_utils.py
```diff
@@ -46,15 +46,10 @@ def get_spanner_destination_uri(self) -> str:
 
     def execute(self, query: str, job_config: Optional[bigquery.QueryJobConfig] = None) -> bigquery.table.RowIterator:
         """Executes a query and returns the result."""
-        if not self.client:
-            logging.error("BigQuery client not initialized")
-            raise RuntimeError("BigQuery client not initialized")
-            
-        start_time = time.time()
-        logging.info(f"Executing query (first 100 chars): {query.strip()[:100]}...")
+        query_job = self.execute_async(query, job_config)
```
SandeepTuniki: Should this statement be inside the `try` block?

import-automation/workflow/ingestion-helper/aggregation_utils.py
```diff
@@ -610,15 +654,32 @@ def run_aggregation(self, import_list: List[Dict[str, Any]]) -> bool:
                     job_config = bigquery.QueryJobConfig(query_parameters=[
                         bigquery.ScalarQueryParameter("import_name", "STRING", import_name),
                     ])
+                    # Note: You might want to make this async too if it's slow, but keeping it sync for now 
+                    # assuming it's just logging execution time.
```
SandeepTuniki: Stray AI-comment? (ignore if manually written).

SandeepTuniki (approved):

import-automation/workflow/ingestion-helper/aggregation_utils.py
vish-cs: Removed

import-automation/workflow/ingestion-helper/aggregation_utils.py
```diff
@@ -63,25 +58,73 @@ def execute(self, query: str, job_config: Optional[bigquery.QueryJobConfig] = No
...
+                elif job.error_result:
+                    overall_status = "FAILED"
+                    failed_jobs.append(job_id)
+                    error_message += f"Job {job_id} failed: {job.error_result}. "
+            except Exception as e:
+                logging.error(f"Failed to get job status for {job_id}: {e}")
+                overall_status = "FAILED"
+                failed_jobs.append(job_id)
+                error_message += f"Failed to get job {job_id}: {e}. "
+
```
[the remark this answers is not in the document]
vish-cs: Updated

import-automation/workflow/spanner-ingestion-workflow.yaml
```diff
@@ -66,12 +66,35 @@ main:
                 call: http.post
                 args:
                   url: ${helper_url}
-                  timeout: 1800
                   auth:
```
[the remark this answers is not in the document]
vish-cs: Added a 5 minute timeout

import-automation/workflow/ingestion-helper/aggregation_utils.py
vish-cs: Good call..done

gmechali:
Putting this here but we already discussed in chat - pls dont submit this. It would break the DCP setup.
Can you modify this PR such that the async mode is tied to an environment variable or a flag on the ingestion helper, so that you can enable it for base DC, but keep it disabled for DCP, until we can fix DCP to support it (which means workflow change + terraform change).

In general we should always strive to keep backwards compatibility, move the traffic over, then tear down the old code. We've been getting away with not doing that because it was just our instance, but this is definitely changing now :)

vish-cs:
> Putting this here but we already discussed in chat - pls dont submit this. It would break the DCP setup. Can you modify this PR such that the async mode is tied to an environment variable or a flag on the ingestion helper, so that you can enable it for base DC, but keep it disabled for DCP, until we can fix DCP to support it (which means workflow change + terraform change).
> 
> In general we should always strive to keep backwards compatibility, move the traffic over, then tear down the old code. We've been getting away with not doing that because it was just our instance, but this is definitely changing now :)

I have updated the code to support both sequential and parallel execution of aggregation queries using is_base_dc flag. Will proceed with submission to unblock base DC ingestion.

PR #2035, opened 2026-05-24
[DCP - Ingestion] Allow Ingestion Helper to start clearing the redis cache

gmechali:
This needs to be triggered by the workflow at the end of every workflow

### DIFF
--- a/import-automation/workflow/ingestion-helper/main.py
+++ b/import-automation/workflow/ingestion-helper/main.py
@@ -289,5 +289,23 @@ def ingestion_helper(request):
         except Exception as e:
             return (f"Aggregation failed: {str(e)}", 500)
 
+    elif action_type == 'clear_redis_cache':
+        logging.info("Action: clear_redis_cache")
+        redis_host = os.environ.get("REDIS_HOST")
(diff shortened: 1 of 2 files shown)


import-automation/workflow/ingestion-helper/main.py
```diff
@@ -289,5 +289,23 @@ def ingestion_helper(request):
         except Exception as e:
             return (f"Aggregation failed: {str(e)}", 500)
 
+    elif action_type == 'clear_redis_cache':
```
vish-cs, line 292: Can we move the logic into a helper file (redis_client.py) and invoke the helper function so the main.py can stay cleaner.

vish-cs (approved):

import-automation/workflow/ingestion-helper/main.py
gmechali: Yes absolutely, I made. a major refactor here:  https://github.com/datacommonsorg/data/pull/2036
Didnt want to slow down this fix for our demo, but] yes I do agree the refactor is needed. PTAL at that PR!

PR #2038, opened 2026-05-25
dc baseline

ajaits:
- **Support 0 eval results**
- **Use environment variable for DC API root**
- **cleanup dc_api_root configs**
- **lint fix**
- **scripts for baseline metrics**


tools/agentic_import/metrics/pvmap_generator_metrics.py
```diff
@@ -0,0 +1,245 @@
...
+          stats: output stats dictionary into which stats are added.
+
+        """
+        if stats is None:
+            stats = dict({
+                'false_positive': 0,
+                'false_negative': 0,
+                'true_positive': 0,
+            })
+        counters = diff_stats.get('counters', {})
```
ajaits, line 188: F1 score from node diffs

PR #2044, opened 2026-05-28
changed rule to new rule Bls_cpi_Category

niveditasing:
This PR includes: 
1. Updated the threshold rule ID. This import previously had a threshold of 1230 with rule id that checks  deletions in numbers, but because the rule was recently changed to check the deletion threshold as a percentage rather than a raw number, the import is currently failing due to deletions.

Fix: Added the relevant rule ID and set the threshold to 0.6% (representing 1,230 rows out of 211,256). 

_`1230/211256 x 100  ~ 0.58223% ~ 0.6%`_ 
Keeping the threshold at this level ensures we capture a meaningful baseline of the data without letting minor anomalies skew the results.

2. Pvmap fix as per new Data additions 

3. The Problem
  The CI build (python_install step in Cloud Build) recently started failing with:

   1 pip._vendor.resolvelib.resolvers.ResolutionTooDeep: 200000
  Because dependencies in requirements_all.txt are unpinned, standard pip3 attempts to fetch the latest versions. Recent upstream updates on
  PyPI (within the google-cloud-* / protobuf ecosystems) introduced a complex version conflict. Standard pip3's backtracking resolver got
  stuck in an infinite loop trying to reconcile these versions, exceeding its 200,000-round limit and crashing.

 Fix:
  Updated the setup_python function in run_tests.sh to:
   1. Bootstrap uv: Check if uv is installed; if not (as is the case in the raw Cloud Build container), install it into the virtual
      environment using pip3 (takes < 1 second).
   2. Install via uv: Use uv pip install instead of standard pip3 to install requirements.

  Why it works: uv is powered by a high-performance Rust resolver (PubGrub) that handles deep backtracking effortlessly. It resolves and installs the entire unpinned dependency graph in under 2 seconds.


statvar_imports/us_bls/cpi_category/validation_config.json
```diff
@@ -2,10 +2,10 @@
     "schema_version": "1.0",
     "rules": [
         {
-            "rule_id": "check_deleted_records_count",
-            "validator": "DELETED_RECORDS_COUNT",
+            "rule_id": "check_deleted_records_percent",
+            "validator": "DELETED_RECORDS_PERCENT",
             "params": {
-                "threshold": 1230
+                "threshold": 0.1
```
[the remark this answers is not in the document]
niveditasing: not needed

saanikaaa (approved):

statvar_imports/us_bls/cpi_category/validation_config.json
```diff
@@ -2,10 +2,10 @@
     "schema_version": "1.0",
     "rules": [
         {
-            "rule_id": "check_deleted_records_count",
-            "validator": "DELETED_RECORDS_COUNT",
+            "rule_id": "check_deleted_records_percent",
+            "validator": "DELETED_RECORDS_PERCENT",
             "params": {
-                "threshold": 1230
+                "threshold": 0.6
```
hareesh-ms, line 8: Since we are adding a threshold, can we ensure to add the golden checks to ensure that the important place,SV combinations are not deleted.

hareesh-ms (changes requested):

statvar_imports/us_bls/cpi_category/validation_config.json
niveditasing: Added golden_summary_report.csv, which captures the important SVs, and  NumPlaces column taking care of the places.

statvar_imports/us_bls/cpi_category/validation_config.json
```diff
@@ -2,10 +2,17 @@
...
+            }
+        },
+        {
+            "rule_id": "check_goldens_summary_report",
+            "validator": "GOLDENS_CHECK",
+            "params": {
+                "golden_files": "../../../../golden_data/golden_summary_report.csv"
```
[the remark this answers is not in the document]
niveditasing: not needed path should be relative only

hareesh-ms (approved):
Lets revisit the logic to chose golden

statvar_imports/us_bls/cpi_category/validation_config.json
```diff
@@ -2,10 +2,17 @@
     "schema_version": "1.0",
     "rules": [
         {
-            "rule_id": "check_deleted_records_count",
-            "validator": "DELETED_RECORDS_COUNT",
+            "rule_id": "check_deleted_records_percent",
+            "validator": "DELETED_RECORDS_PERCENT",
```
pulkeet-wq, line 6: golden checks for places are missing. please add the same

balit-raibot: same has been added in this PR: https://github.com/datacommonsorg/data/pull/2090

PR #2047, opened 2026-05-28
[DCP - Ingestion] Match the ingestion path 

gmechali:
If there is no env var specified, this would be a no-op ensuring this does not change anything for Base DC. 
This PR goes in tandem with: https://github.com/datacommonsorg/datacommons/pull/92


clincoln8 (approved):
Thanks!

PR #2049, opened 2026-05-29
update the timeout for running embeddings

shixiao-coder:


import-automation/workflow/ingestion-helper/embedding_utils.py
```diff
@@ -43,14 +43,15 @@ def get_latest_lock_timestamp(database):
         raise
     return None
 
-def get_updated_nodes(database, timestamp, node_types):
+def get_updated_nodes(database, timestamp, node_types, timeout=None):
```
gmechali: can you just set thed efault here instead of inline with the call on l82?

gmechali: same here

import-automation/workflow/ingestion-helper/main.py
```diff
@@ -45,11 +45,13 @@
     'is_base_dc',
     os.environ.get('IS_BASE_DC', 'true').lower() == 'true',
     'Is base DC')
+flags.DEFINE_integer('timeout', 1700, 'Timeout for Spanner queries')
```
gmechali: Can do what the other flags are doing here, and make the default be os.environ.get(TIMEOUT, 1700). This way we can update it in the terraform.

Also can you be a bit more descriptive on "spanner queries".

import-automation/workflow/ingestion-helper/embedding_utils.py
shixiao-coder: done

import-automation/workflow/ingestion-helper/main.py
shixiao-coder: updated to read from env variable

import-automation/workflow/ingestion-helper/embedding_utils.py
```diff
@@ -43,14 +43,15 @@ def get_latest_lock_timestamp(database):
         raise
     return None
 
-def get_updated_nodes(database, timestamp, node_types):
+def get_updated_nodes(database, timestamp, node_types, timeout=300):
```
gmechali: why do some of these default to timeout=300 but others timeout=timeout? 

Can we make them all use the flag? 
Should we also make the flag be a bit more descriptive? is it a timeout just for spanner queries?

gmechali: The flag already has a default now, so that's why I think the 300 hardcoded default is no longer necessary. 
We could set the default in the flag definition to 300 though if you think it's a better default.

shixiao-coder: the 300 is the source function and the timeout=timeout is passing the params into the spanner call, let me delete the 300, I think we should just let all read from env var from main instead of having internal default value, also updating the flags

shixiao-coder: updated

gmechali (approved):

PR #2072, opened 2026-06-17
US_EIA Fix import path

kartik-s21:
This PR fixes a ModuleNotFoundError: No module named 'util' error that blocked the execution of us_eia scripts (both locally and when running in cloud environments).
 The issue arose because of a recent change in the shared util/ package, which broke the legacy directory-appending strategy used by the EIA ingestion scripts.


scripts/us_eia/opendata/process/common.py
```diff
@@ -26,6 +26,7 @@
 _MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
 _SCRIPT_PATH = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.join(_SCRIPT_PATH, '../../../../util/'))
+sys.path.append(os.path.join(_SCRIPT_PATH, '../../../../'))
```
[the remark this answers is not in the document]
kartik-s21: Done.

balit-raibot (approved):

scripts/us_eia/opendata/process/common.py
```diff
@@ -25,7 +25,7 @@
 # Setup path for import from data/util
 _MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
 _SCRIPT_PATH = os.path.dirname(os.path.abspath(__file__))
-sys.path.append(os.path.join(_SCRIPT_PATH, '../../../../util/'))
+sys.path.append(os.path.join(_SCRIPT_PATH, '../../../../'))
```
vish-cs, line 28: I did not realize that it could break the import scripts. I wonder if it can impact other imports as well. Instead of fixing various imports, can we revert the changes to file_util.py in my PR and instead handle it in differ_utils.py

-from util.file_util import FileIO
-from util.file_util import file_get_matching
+import sys
+_SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+_DATA_DIR = os.path.dirname(os.path.dirname(_SCRIPT_DIR))
+sys.path.append(os.path.join(_DATA_DIR, 'util'))
+
+from file_util import FileIO
+from file_util import file_get_matching

kartik-s21: yes sure this will be fine.

PR #2084, opened 2026-07-07
Config changes us hud income limits

shvngisingh:
Made changes in downloading script to get the recent date.
Logic:
Resolving the range() function which has stop value, so Python terminates and does not execute the loop for this value .So skipping the recent year.


scripts/us_hud/income/process.py
```diff
@@ -146,15 +146,28 @@ def process_all():
 
     if FLAGS.mode == "" or FLAGS.mode == "download":
         logging.info("Starting download phase...")
-        for year in range(2006, today.year):
+        for year in range(2006, today.year + 1):
             url = get_url(year)
             if url:
                 filename = f"Section8-FY{year}.xlsx" if year > 2016 else f"Section8-FY{year}.xls"
+                if year == today.year:
```
pulkeet-wq, line 153: WHy we need these extra lines of code for processing the data of the current year? Please add proper comments against the code

shvngisingh: As per Gemini comments : Including the current year (today.year) in the download loop is a good addition to fetch the latest data. However, HUD income limits for the current year are typically not released until April or May. During the first few months of any year, attempting to download the current year's file will result in a 404 error. To resolve this issue added the lines

scripts/us_hud/income/validation_config.json
```diff
@@ -0,0 +1,29 @@
...
+        },
+        {
+            "rule_id": "check_goldens_summary_report",
+            "validator": "GOLDENS_CHECK",
+            "params": {
+                "golden_files": "../../../../golden_data/golden_summary_report.csv"
+            }
+        },
+        {
+            "rule_id": "Check_goldens_output_csv",
```
pulkeet-wq: please change the rule I'd to snake case

shvngisingh: changed

hareesh-ms (approved):

balit-raibot (approved):

PR #2112, opened 2026-07-24
Switch autorefresh to prod DC API

ajaits:
Have import automation use prod DC API for dc-import tool as well as other calls.


saanikaaa (approved):

ajaits:
dc-import tool fixed in https://github.com/datacommonsorg/import/pull/691.
This is no longer required.

PR #2135, opened 2026-08-05
Differ updates and fixes

vish-cs:
This PR makes the following changes to the differ:
a) Added a new output file containing the original values of modified nodes (for debugging)
b) Updated MCF parsing logic to handle quoted strings cleanly
c) Fixes the differ counter names to check for SKIP state in the executor

### DIFF
--- a/tools/import_differ/differ_utils.py
+++ b/tools/import_differ/differ_utils.py
@@ -9,55 +9,24 @@ import sys
 _SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
 _DATA_DIR = os.path.dirname(os.path.dirname(_SCRIPT_DIR))
 sys.path.append(os.path.join(_DATA_DIR, 'util'))
+sys.path.append(os.path.join(_DATA_DIR, 'tools', 'statvar_importer'))
 
 from file_util import FileIO
 from file_util import file_get_matching
-
-_MCF_VALUE_SPLIT_RE = re.compile(r'\s*(?:"[^"\\]*(?:\\.[^"\\]*)*"|[^,]+)')
-
-
-def _parse_mcf_value(val: str):
-    val = val.strip()
-    if ',' not in val:
-        return val
-    parts = [p.strip() for p in _MCF_VALUE_SPLIT_RE.findall(val) if p.strip()]
-    return parts if len(parts) > 1 else val
+from mcf_file_util import load_mcf_nodes
 
 
-def load_mcf_file(file: str):
+def load_mcf_file(file: str) -> list:
     """ Reads an MCF text file and returns mcf nodes."""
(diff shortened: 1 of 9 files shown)


tools/import_differ/differ_utils.py
```diff
@@ -28,7 +28,14 @@ def load_mcf_file(file: str):
...
             parsed_line = mcf_line.match(line)
             if parsed_line is not None:
-                current_mcf_node[parsed_line.group(1)] = parsed_line.group(2)
+                prop = parsed_line.group(1)
+                val = parsed_line.group(2).strip()
+                # Parse unquoted comma-separated lists (e.g. dcid:A, dcid:B) into a Python list
+                if ',' in val and not (val.startswith('"') and val.endswith('"')):
+                    parts = [p.strip() for p in val.split(',')]
+                    if all(' ' not in p for p in parts):
+                        val = parts
```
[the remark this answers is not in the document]
vish-cs: Updated to use a regex

tools/import_differ/differ_utils.py
```diff
@@ -13,6 +13,16 @@
...
 from file_util import file_get_matching
 
+_MCF_VALUE_SPLIT_RE = re.compile(r'\s*(?:"[^"\\]*(?:\\.[^"\\]*)*"|[^,]+)')
+
+
+def _parse_mcf_value(val: str):
+    val = val.strip()
+    if ',' not in val:
+        return val
+    parts = [p.strip() for p in _MCF_VALUE_SPLIT_RE.findall(val) if p.strip()]
```
ajaits: Does the regex handle escaped double quotes within?
https://github.com/datacommonsorg/data/blob/master/tools/statvar_importer/mcf_file_util.py#L733 is also an option here.

tools/import_differ/import_differ.py
```diff
@@ -79,8 +79,10 @@
 
 def val_str(value) -> str:
     if isinstance(value, list):
-        return ",".join([val_str(v) for v in value])
-    if value and isinstance(value, str) and " " in value and value[0].isalpha():
+        return ", ".join([val_str(v) for v in value])
```
ajaits: may want to normalize before comparison:
- return list in sorted order
- normalize or stip namespace

Here is an example: https://github.com/datacommonsorg/data/blob/master/tools/statvar_importer/mcf_file_util.py#L909

tools/import_differ/differ_utils.py
vish-cs: I have updated the mcf parsing logic to use the existing functions in mcf_file_util

tools/import_differ/import_differ.py
vish-cs: I have updated the mcf converstion logic to use the existing functions in mcf_file_util

ajaits (approved):

PR #2142, opened 2026-08-06
Fix Pandas 3.x compatibility, float-casting, and test discovery gap for USCensusPEP_Sex

rohitkumarbhagat:
#### Context & Problem
The scheduled execution for the `USCensusPEP_Sex` import failed in production due to the following reasons:
1. **Pandas 3.x Deprecation Removal**: The preprocessing script called `pd.read_table(..., delim_whitespace=True)`. This keyword was completely removed in Pandas 3.x, causing the pipeline to crash in the production environment.
2. **Silent Test Discovery Skipping**: Unit tests for this import were silently skipped in CI builds because the directories `scripts/us_census/pep/` and `scripts/us_census/pep/us_pep_sex/` were missing package-initialization `__init__.py` files, masking errors.
3. **Float-Casting Bug**: Re-enabling the tests revealed a bug where a blank row in county intercensal files introduced `NaN` values, forcing year and FIPS codes to be cast to floats (resulting in malformed values like `1989.0` and `geoId/1001.0` in the CSV outputs).

#### Solution & Changes
1. **Pandas 3.x Compatibility**:
   * Replaced all instances of `delim_whitespace=True` with `sep=r'\s+'` in [`scripts/us_census/pep/us_pep_sex/process.py`](file:///usr/local/google/home/rohitrkumar/Documents/dc/github/rohitkumarbhagat/data-1/scripts/us_census/pep/us_pep_sex/process.py).
   * Added checks to guard `absl.flags` definitions in `process.py` from duplicate definition crashes during test discovery.

2. **Data Type and Format Correctness**:
   * Updated the intercensal county parsing function to drop NaN/blank rows immediately and cast year/FIPS columns explicitly to `int64`. This fixes the malformed `1989.0` years and `geoId/1001.0` county codes.
   * Cast the melted `Observation` column explicitly to `int64` to prevent `.0` suffixes in the final CSV file.
   * Regenerated the expected unit test CSV, MCF, and TMCF files to align with the correct integer-formatted output.

3. **CI/CD Enablement & Code Cleanup**:
   * Created empty package `__init__.py` files in both PEP directory levels to ensure unit tests are auto-discovered and executed by CI.
   * Refactored the adjacent `annual_population` unit test [`preprocess_test.py`](file:///usr/local/google/home/rohitrkumar/Documents/dc/github/rohitkumarbhagat/data-1/scripts/us_census/pep/annual_population/preprocess_test.py) to write outputs to a temporary directory rather than writing directly into the local git workspace.

#### Verification
* All PEP tests auto-discovered and **passed successfully** (`OK`).
* Project linter check passed successfully with zero warnings/errors.


scripts/us_census/pep/annual_population/preprocess_test.py
```diff
@@ -42,24 +42,22 @@ def __init__(self, methodName: str = ...) -> None:
...
+            with open(tmcf_path, encoding="UTF-8") as tmcf_file:
+                self._actual_tmcf_data = tmcf_file.read()
 
-        with open(tmcf_path, encoding="UTF-8") as tmcf_file:
-            self._actual_tmcf_data = tmcf_file.read()
-
-        with open(cleaned_csv_path, encoding="utf-8") as csv_file:
-            self._actual_csv_data = csv_file.read()
+            with open(cleaned_csv_path, encoding="utf-8") as csv_file:
+                self._actual_csv_data = csv_file.read()
```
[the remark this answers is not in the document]
rohitkumarbhagat: done

scripts/us_census/pep/us_pep_sex/process.py
```diff
@@ -1092,6 +1101,7 @@ def process(self):
                 value_vars=['Count_Person_Male', 'Count_Person_Female'],
                 var_name="SV",
                 value_name="Observation")
+            final_df['Observation'] = final_df['Observation'].astype('int64')
```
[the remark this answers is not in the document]
rohitkumarbhagat: done

scripts/us_census/pep/annual_population/preprocess.py
```diff
@@ -54,10 +54,12 @@
 
 _FLAGS = flags.FLAGS
 
-flags.DEFINE_string('mode', '', 'Options: download or process')
-flags.DEFINE_bool(
-    'is_summary_levels', False,
-    'Options: True for all summary_levels and False for only 162')
+if 'mode' not in flags.FLAGS:
```
ajaits: Is there a redefinition or cyclic dependency with this?

rohitkumarbhagat: Yes, this prevents a flag redefinition rather than a cyclic dependency. Abseil uses a process-global flag registry, and test discovery imports both `annual_population.preprocess` and `us_pep_sex.process`; both define the same mode string flag. Without the guard, whichever module is imported second raises DuplicateFlagError.

rohitkumarbhagat: i have refactored code to lazy load the flags, to avoid cases where redeclaration conflict.

scripts/us_census/pep/annual_population/preprocess_test.py
```diff
@@ -21,11 +21,19 @@
...
+_SCRIPTS_DIR = os.path.abspath(os.path.join(_MODULE_DIR, '../../../'))
+if _SCRIPTS_DIR not in sys.path:
+    sys.path.insert(0, _SCRIPTS_DIR)
 sys.path.insert(1, _MODULE_DIR)
+
 # pylint: disable=wrong-import-position
 # pylint: disable=import-error
-from preprocess import process
-from constants import TEST_DATA_DIR
+try:
```
ajaits: havn't seen this pattern of 'try..except' for imports. we have instead used sys.path.append() to add multiple folders so all imports can be done without '.' and script can be run frm any folder too

ajaits (approved):

scripts/us_census/pep/annual_population/preprocess_test.py
rohitkumarbhagat: Good point. Removed the `try/except` fallback and added both the scripts and module directories to `sys.path`. The package-qualified imports now work with recursive test discovery and when running the test directly from its directory. Applied the same cleanup to the analogous `us_pep_sex` test as well.

PR #2159, opened 2026-08-14
Add GCS download mode for Tennessee diabetes import

rohitkumarbhagat:
## Summary

  - Add `--download_source=tn|gcs` to the Tennessee diabetes downloader.
  - Keep TN.gov as the default for direct execution.
  - ~Enable GCS mode in the import manifest.~ - last download from TN.gov passed. Keeping it default.
  - Generate yearly filenames from 2019 through the current year for both modes.
  - Skip missing GCS years, fail when no files are downloaded, and propagate unexpected GCS errors.
  - Log download source, destination, completion, and file size.
  - Document the semi-manual GCS staging workflow.

  ## Motivation

  TN.gov blocks requests originating from Google Cloud environments. GCS staging provides a semi-manual fallback while preserving the existing processing flow.

  ## Validation

  - Added four focused unit tests for URL generation and GCS behavior.
  - All focused tests pass.
  - Repository lint passes.
  - Completed a live GCS download and Excel-processing validation for the 2019–2022 files.

  The runtime service account requires `storage.objects.get` access to the `unresolved_mcf` bucket.


statvar_imports/nyu_diabetes/tennessee/download.py
```diff
@@ -55,14 +70,56 @@ def download_files(url_list, save_folder):
...
+                parsed_url.path.lstrip('/'))
+            blob.download_to_filename(file_path)
+        except exceptions.NotFound:
+            Path(file_path).unlink(missing_ok=True)
+            logging.warning(f"GCS object not found, skipping: source={url}")
+            continue
+        except Exception as e:
+            Path(file_path).unlink(missing_ok=True)
+            e.add_note(f"Failed to download GCS object {url}")
+            raise
```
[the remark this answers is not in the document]
rohitkumarbhagat: script would run in python 3.11+

statvar_imports/nyu_diabetes/tennessee/download.py
```diff
@@ -55,14 +70,56 @@ def download_files(url_list, save_folder):
...
+        logging.info(
+            f"Starting download: source={url}, destination={file_path}")
+
+        try:
+            blob = storage_client.bucket(parsed_url.netloc).blob(
+                parsed_url.path.lstrip('/'))
+            blob.download_to_filename(file_path)
+        except exceptions.NotFound:
+            Path(file_path).unlink(missing_ok=True)
+            logging.warning(f"GCS object not found, skipping: source={url}")
```
ajaits, line 103: Should this be fatal so script stops if GCS file is missing?

statvar_imports/nyu_diabetes/tennessee/download.py
```diff
@@ -110,12 +167,17 @@ def process_excel_files(input_dir):
     logging.info("\nExcel file processing complete. 'year' column added to all processed files.")
 
 def main(_):
-    url_template ="https://www.tn.gov/content/dam/tn/health/documents/vital-statistics/death/{}/Diabetes_County_{}.xlsx"
+    tn_url_template = "https://www.tn.gov/content/dam/tn/health/documents/vital-statistics/death/{year}/Diabetes_County_{year}.xlsx"
+    gcs_url_template = "gs://unresolved_mcf/nyu_diabetes/tennessee/latest/input_files/Diabetes_County_{year}.xlsx"
     start_year = 2019
     current_year = date.today().year
```
ajaits, line 173: if source doesn't publish new year's data on Jan 1, can set this to be a month or two behind so run in Jan/Feb will only require prev year's data.

ajaits (approved):

statvar_imports/nyu_diabetes/tennessee/download.py
rohitkumarbhagat: I mimicked the current behavior where only available files are processed. It would fail if no files are available.  At present data is present only till 2022, hence import would fail if file existence is enforced.

statvar_imports/nyu_diabetes/tennessee/download.py
rohitkumarbhagat: I’d prefer to keep current_year as-is. Files that have not yet been published are skipped, so trying the current year does not block the import. It also lets us pick up newly published data immediately without waiting for a code change or an artificial month offset.

PR #2187, opened 2026-08-26
Add logic to copy historical data

vish-cs:


import-automation/executor/app/executor/import_executor.py
```diff
@@ -844,6 +844,70 @@ def _update_latest_version(self, version, output_dir, import_spec,
...
+                        blob.name.endswith('/')):
+                    continue
+                if blob.name.startswith(prefix_with_slash):
+                    rel_path = blob.name[len(prefix_with_slash):]
+                else:
+                    rel_path = os.path.basename(blob.name)
+                if not rel_path or rel_path.endswith('/'):
+                    continue
+                if rel_path.startswith('genmcf/'):
+                    rel_path = rel_path[len('genmcf/'):]
```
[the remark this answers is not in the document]
vish-cs: Done

import-automation/executor/app/executor/import_executor.py
```diff
@@ -888,11 +952,13 @@ def _import_one_helper(
...
-                tmpdir,
-                timeout=self.config.venv_create_timeout,
-            )
+            interpreter_path = sys.executable
+            process = subprocess.CompletedProcess(args=[], returncode=0)
+            # interpreter_path, process = _create_venv(
+            #     [central_requirements_path, requirements_path],
+            #     tmpdir,
+            #     timeout=self.config.venv_create_timeout,
+            # )
```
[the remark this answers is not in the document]
vish-cs: done

import-automation/executor/app/executor/import_executor.py
```diff
@@ -844,6 +844,62 @@ def _update_latest_version(self, version, output_dir, import_spec,
...
+            logging.info(
+                'Skipping GCS historic data ingestion due to skip_gcs_upload=True'
+            )
+            return
+
+        try:
+            storage_client = storage.Client(self.config.gcs_project_id)
+            bucket = storage_client.bucket(
+                self.config.storage_prod_bucket_name)
+            source_prefix = f'{output_dir}/historic/'
```
ajaits: Wondering if it would be safer to list specific files in GCS to be used in the manifest.json rather than picking up all files. This would make reviews easier, o/w copy to historic folder will have to wait until the update is reviewed.

On the flip side, this will add some overhead of managing the list through manual PRs rather than automatic simple file copy.

import-automation/executor/app/executor/import_executor.py
```diff
@@ -844,6 +844,62 @@ def _update_latest_version(self, version, output_dir, import_spec,
...
+                if blob.size:
+                    historic_data_size += blob.size
+        except Exception as e:
+            logging.error(f'Failed to copy historic data from GCS: {e}')
+            raise
+
+        if import_summary:
+            import_summary.import_stats['historic_data_size'] = historic_data_size
+            import_summary.data_volume += historic_data_size
+        logging.info(
```
ajaits, line 906: Can we also log the number and names of files that were copied?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -943,6 +999,12 @@ def _import_one_helper(
                 import_summary.import_stats.get('validation_data_size', 0))
             logging.info(import_summary)
 
+            if self.config.ingest_historic_data:
```
ajaits, line 1009: Can this be a list of files rather than a flag? If non-empty, then files are copied.

import-automation/executor/app/executor/import_executor.py
ajaits: instead of hardcoding the folder '/historic/', can this be a parameter that defaults to /historic/?

import-automation/executor/app/executor/import_executor.py
```diff
@@ -844,6 +844,62 @@ def _update_latest_version(self, version, output_dir, import_spec,
...
+                    rel_path = rel_path[len('genmcf/'):]
+                dest_blob_name = f'{output_dir}/{version}/historic/genmcf/{rel_path}'
+                logging.info(
+                    f'Copying historic GCS blob {blob.name} to {dest_blob_name}'
+                )
+                bucket.copy_blob(blob, bucket, dest_blob_name)
+                if blob.size:
+                    historic_data_size += blob.size
+        except Exception as e:
+            logging.error(f'Failed to copy historic data from GCS: {e}')
```
ajaits, line 899: pls log the src and dst file names as well.

ajaits (approved):

import-automation/executor/app/executor/import_executor.py
vish-cs: Added the number. We log each file copied separately.

import-automation/executor/app/executor/import_executor.py
vish-cs: we already do that
         logging.info(
                    f'Copying historic GCS blob {blob.name} to {dest_blob_name}'
                )

import-automation/executor/app/executor/import_executor.py
vish-cs: Change to make it a function argument.

vish-cs: Agree. That's the idea to keep it simple for now. Allows us to iterate over this quickly especially through the planned ET migration. We can start with a simple flag and then update as needed.

import-automation/executor/app/executor/import_executor.py
vish-cs: Leaning toward a flag as mentioned above

PR #2192, opened 2026-08-31
Centralize differ DataFrame column definitions

rohitkumarbhagat:
Addresses the follow-up cleanup for DuckDB differ registration:

  - Defines DIFFER_COLUMNS as the shared differ schema.
  - Initializes the runner’s empty differ DataFrame with that schema.
  - Reuses the schema when loading empty differ output and registering an empty DataFrame with DuckDB.
  - Adds regression coverage for missing and empty differ data.

  All 121 tools/import_validation tests pass.


vish-cs (approved):
Thanks for the clean up!

